PR #20, opened 2021-08-02
Implement URL permalinks & improve <title> tag

### LINKED ISSUE #19: Implement URL permalinks & improve <title> tag
URL-based routing can be used so that people can share links to specific movies

JoshHeng:
Fixes #19

### DIFF
--- a/src/views/Movie.js
+++ b/src/views/Movie.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { useRouteMatch, useHistory } from 'react-router-dom'
 import { Title } from '../components/Title'
 import { Card } from '../components/Card'
 import { useMovie } from '../hooks/useMovie'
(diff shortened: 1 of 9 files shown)


src/views/Movie.js
```diff
@@ -9,41 +11,57 @@ import { getStreamUrl } from '../lib/index'
 import './Movie.css'
 
 export function MovieView(props) {
+    const baseRouteMatch = useRouteMatch('/:type/:source/:title/:slug');
```
JamesHawkinss, line 14: I reckon we could just remove :title from the route and have `/:type/:source/:slug` and similar for shows. Is there a specific reason that you're including title in the route?

src/views/Search.js
```diff
@@ -107,22 +114,41 @@ export function SearchView() {
...
             })
         }
         fetchHealth()
-    }, [])
+    }, []);
+
+    React.useEffect(() => {
+        if (streamRouteMatch) {
+            if (streamRouteMatch?.params.type === 'movie' || streamRouteMatch.params.type === 'show') {
+                if (streamRouteMatch?.params.source === 'lookmovie' || streamRouteMatch.params.source === 'gomostream') {
```
JamesHawkinss, line 126: I would avoid hardcoding the possible sources here. Adds an extra layer of complication when adding new sources in the future, and could lead to issues if a developer forgets to add their source to the if statement.

src/views/Search.js
```diff
@@ -107,22 +114,41 @@ export function SearchView() {
...
+                    getStream(streamRouteMatch.params.title, streamRouteMatch.params.slug, streamRouteMatch.params.type, streamRouteMatch.params.source);
+                }
+                else return setErrorStatus(`I couldn't find that ${streamRouteMatch.params.type}. Please try searching below.`);
```
JamesHawkinss, line 129: This is very nitpicky (sorry) but could you ditch the first-person errors? Doesn't match the rest of the error messages :p

JamesHawkinss (changes requested):
Looks fantastic! Just a few small things.

src/views/Movie.js
JoshHeng: :title is used for the title that appears on the top of the Video page (+ in the <title> tag). The alternative to this could be to display the slug instead, or to not display the title at all?

JamesHawkinss: Given we have the slug on the movie page, it might be possible to fetch information about the show (such as the title) from Movie.js - I see what you mean though. Leave the title in for now, it can always be changed in the future.

PR #24, opened 2021-08-06
Continue watching

### LINKED ISSUE #8: Save last watched episode/season
Store last watched episode/season in localstorage, and if present load from localstorage.

JamesHawkinss:
Implements the continue watching feature. Based on code from @JipFr. 

Fixes #8


JoshHeng (changes requested):
When loading a TV show, the episode selector doesn't work & the video doesn't load.
This seems to be an issue with making the episode selector responsive (which is also an issue in prod), however in this PR the list of episodes is not shown either

JoshHeng (changes requested):
If I'm playing a TV show, clicking a new season to preview the episodes will refresh the video/skip it back by like 0.5 seconds.

PR #27, opened 2021-08-30
Add Cloudflare worker CORS proxy

JoshHeng:


.env
```diff
@@ -0,0 +1 @@
+REACT_APP_CORS_PROXY_URL=https://PROXY.workers.dev?destination=
```
JamesHawkinss, line 1: Could you update the default value to `https://proxy-01.no-cors-proxy.workers.dev`?

PR #31, opened 2021-09-26
fix: Fix lookmovie security bug by passing PHPSESSID cookie & user agent

JoshHeng:


JoshHeng:
The proxy will also need redeploying with the new script

PR #45, opened 2021-11-11
"react-i18next" infrastructure created

05akalan57:
In this commit, language support and Turkish language were added using the "react-i18next" framework (the site opens that way according to the browser language)


JamesHawkinss:
Thanks for the contribution, but v2 has been created and will not be compatible with these changes.

PR #51, opened 2022-01-03
Feature - add keyboard navigation support

sammce:
This pull requests allows users to navigate throughout the site using the keyboard. Navigating using the tab key is a feature which is used extensively by lesser-abled users, for instance those with poor fine-motor skills who have trouble using a mouse. Without this, anyone who fits this description is unable to use the site.

I have outlined what the site looked like before and after the changes in the following videos.

https://user-images.githubusercontent.com/78268167/147933350-1c98507f-d8ba-4db8-8751-510b7be9eeeb.mp4

I'm attempting to switch to TV shows without using the mouse, which was not possible before this pr.

The changes I've made simply involve the built in HTML property `tabindex` ([Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex)). I've also added a border style whenever the required elements are focused with tab using the CSS pseudo-selector `:focus-visible` ([Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible))

The benefit of using `:focus-visible` is that the styles are only applied when focused with the keyboard, so there will be no noticeable difference for regular users, but the functionality will be there for those who need it.

Below is a visualization of how the site with this pr can now be used with only the tab key.

https://user-images.githubusercontent.com/78268167/147934080-c9ed7540-e736-4e30-8618-165027ab6d4d.mp4

### DIFF
--- a/src/components/MovieRow.css
+++ b/src/components/MovieRow.css
@@ -60,6 +60,15 @@
     transform: translateX(.3rem) translateY(.1rem);
 }
 
+.movieRow:focus-visible {
+    border: 1px solid #fff;
+    background-color: var(--content-hover);
+}
+
+.movieRow:focus-visible .watch .arrow {
+    transform: translateX(.3rem) translateY(.1rem);
+}
+
 .attribute {
     color: var(--text);
     background-color: var(--theme-color);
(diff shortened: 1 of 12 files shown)


JamesHawkinss (approved):
lgtm! thanks for the contribution!

PR #120, opened 2023-01-21
Add german translation

zisra:


zisra:
This one should be OK

mrjvs:
Hey, massive thanks for the translations. We'd love to add German but we currently making some major changes involving lots of new texts.

So because of the new version we've frozen updates for the current version. I'll have to reject this PR because of it.

We still want to add German tho, can we ask you for translations once the new version is ready?

zisra:
Alright, thanks!

PR #127, opened 2023-01-24
Show more episode info in the page title for a series

destruc7i0n:
Was playing with the v3 of the site and thought this might be a good addition, as I generally look at the title for this info

Changes:
- Extracted the logic to generate the S`X`E`Y` from `ShowTitleControl` to a reusable hook
- Created a control to update the title

Just a few questions/points of discussion before a possible merge:
* Better name for `episodeIdentifier`?
* Better way to block usage of data in `useCurrentSeriesEpisodeInfo` to only when it is a series? I currently attempted a way similar to the union pattern I saw used
* Should the data from the new hook be moved into the player state or another store?

### DIFF
--- a/src/components/video/hooks/useCurrentSeriesEpisodeInfo.ts
+++ b/src/components/video/hooks/useCurrentSeriesEpisodeInfo.ts
@@ -1,0 +1,33 @@
+import { useMemo } from "react";
+import { useVideoPlayerState } from "../VideoContext";
+
+export function useCurrentSeriesEpisodeInfo() {
+  const { videoState } = useVideoPlayerState();
+
+  const { current, seasons } = videoState.seasonData;
+
+  const currentSeasonInfo = useMemo(() => {
+    return seasons?.find((season) => season.id === current?.seasonId);
+  }, [seasons, current]);
+
+  const currentEpisodeInfo = useMemo(() => {
+    return currentSeasonInfo?.episodes?.find(
+      (episode) => episode.id === current?.episodeId
+    );
+  }, [currentSeasonInfo, current]);
+
+  const isSeries = Boolean(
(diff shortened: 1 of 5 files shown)


mrjvs:
Damn, our first proper contribution. Code looks great. Thanks a lot.

> Better name for episodeIdentifier?
I'd go with `humanizedEpisodeId`. because its basically a human readable version of the ID.
and it would prevent confusion with the actual episodeId.

> Better way to block usage of data in useCurrentSeriesEpisodeInfo to only when it is a series
I think the way you do it right is perfectly fine, no need to change it.

> Should the data from the new hook be moved into the player state or another store?
I think it should stay as just a hook, (almost) all data thats needed is in the videoState so no need to put a computed value into the store itself.

once you've updated the name of episodeIdentifier, let me know so I can approve and merge

src/components/video/hooks/useCurrentSeriesEpisodeInfo.ts
```diff
@@ -0,0 +1,33 @@
...
+    );
+  }, [currentSeasonInfo, current]);
+
+  const isSeries = Boolean(
+    videoState.seasonData.isSeries && videoState.seasonData.current
+  );
+
+  if (!isSeries) return { isSeries: false };
+
+  const episodeIdentifier = `S${currentSeasonInfo?.number} E${currentEpisodeInfo?.number}`;
```
mrjvs, line 25: I'd rename to `humanizedEpisodeId` to prevent confusion with `episodeId` referenced in other places in the video player state

mrjvs (changes requested):

mrjvs (approved):
looks great, thanks a lot

PR #133, opened 2023-01-29
Update dockerfile

GiddyGoatGaming:
Updates node from 16.15 to 18.13


mrjvs:
please do not provide version bump PR's. we accepted the first one but we will not accept any more

PR #151, opened 2023-02-20
Quality of life fixes

maxwellward:
- Reduces space below search bar on mobile (desktop design unchanged)
- Fixes weird hover state when hovering over titles in edit mode
- Makes title text fade behind the header, instead of overlapping and looking weird
- Improves mobile video player by:
    - Removing "movie-web" text from brand pill (logo remains) only when loading a video on mobile to give more space to the title
    - Truncating media title when in portrait mode (the full name isn't needed, as you tapped the show so you know what it is, the first word or two is enough to remember what you're watching if you forgot, *and* it shows fully in landscape mode anyway)
    - Shrink popups vertically to fit in the screen when on horizontal mobile devices (this does not fix popups going off the side of the screen, that's still an issue. Popups are still normal size on taller devices)
    - Hiding button text for captions, episodes, and providers on small screens

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ To run this project locally for contributing or testing, run the following comma
 git clone https://github.com/movie-web/movie-web
 cd movie-web
 yarn install
-yarn start
+yarn dev
 ```
 
 To build production files, simply run `yarn build`.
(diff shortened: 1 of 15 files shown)


JipFr (changes requested):

mrjvs (approved):
lgtm, but would like it if @JipFr took one final look before merging

Issue #147, opened 2023-02-19
[Bug]: known bugs for V3


mrjvs:
- [ ] when chromecasting, it doesnt start at the time you were at, but starts at the beginning of the video
- [X] `manifestLoadError`, error happens in netfilm provider when loading any HLS stream
- [X] `Misconfigured instance, missing key: NORMAL_ROUTER`, happens on newly deployed instances.
- [ ] HLS doesnt work with chromecasting
- [ ] Subtitles arent sent through for chromecasting
- [x] mobile touch doesnt show backdrop (only if tapping in the area of the pause button)

mrjvs:
gonna make these into new issue, as its basically one bug left

PR #169, opened 2023-02-22
Development tooling, round robin and better settings

mrjvs:
in this PR:
- /dev page with all kinds of tools to test with (providers, embeds, and static video player)
- unit tests for providers, makes it easy to see what providers still work
- NORMAL_ROUTER setting now has a default that doesnt yell at you if you dont configure it
- proxy url settings is now a comma seperated list, backwards compatible. does round robin on all urls.

### DIFF
--- a/index.html
+++ b/index.html
@@ -25,7 +25,7 @@
     />
 
     <script src="config.js"></script>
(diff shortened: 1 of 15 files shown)


mrjvs:
this also fixes one item on #147

JamesHawkinss (approved):

JamesHawkinss (approved):

PR #199, opened 2023-03-12
flixhq scraping improved

frost768:
FlixHQ supported media types are movies and series and it also provides subtitles. These were not included in the result of the scraping. This pull request adds them into the result object.


src/backend/helpers/streams.ts
```diff
@@ -10,9 +10,11 @@ export enum MWCaptionType {
 
 export enum MWStreamQuality {
   Q360P = "360p",
+  Q540P = "540p",
   Q480P = "480p",
   Q720P = "720p",
   Q1080P = "1080p",
+  QAUTO = "auto",
```
mrjvs, line 17: why was QAuto added? it doesnt seem to be used

mrjvs (changes requested):

mrjvs:
tested, works great. just resolve the one review comment and it can be merged

src/backend/helpers/streams.ts
frost768: it does come from flixhq side so if we ever let users to choose their own preferred quality it is just an another option we can display to them

frost768: ok, removed it.

mrjvs (approved):

PR #224, opened 2023-03-23
Playback speed

zisra:


mrjvs:
I cant find any feature requests for this on github or the discord server,
do you have a specific need for changing playback rate?

zisra:
I used an extension to do this previously because I like to be able to watch certain boring parts of an episode at a quicker speed. Also it is part of many standard video players

src/video/state/providers/castingStateProvider.ts
```diff
@@ -87,6 +87,9 @@ export function createCastingStateProvider(
     togglePictureInPicture() {
       // no picture in picture while casting
     },
+    setPlaybackSpeed(num) {
+      this.setPlaybackSpeed(num);
+    },
```
mrjvs, line 92: Please dont make a recursive call lol.

use LoadRequest with a media object that has the proper playbackrate.

see https://developers.google.com/cast/docs/reference/web_sender/chrome.cast.media.LoadRequest

src/video/state/providers/videoStateProvider.ts
```diff
@@ -228,6 +228,10 @@ export function createVideoStateProvider(
         }
       }
     },
+    setPlaybackSpeed(num) {
+      player.playbackRate = num;
+      state.mediaPlaying.playbackSpeed = num;
+    },
```
mrjvs, line 234: in addition to this, you will also need to listen to the playbackrate event on the video player element, thats in case extensions or native browser features allow the user to change the rate manually. It wont desync if we listen for those events.

see https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/ratechange_event

mrjvs (changes requested):
a few more changes and then itll be ready for merge

zisra:
Please don't merge yet, I'm making one more change

zisra:
Here we go

frost768:
Could you use yarn instead of npm? In readme it states "note: must use yarn to install packages and run NodeJS 16"

src/video/state/providers/videoStateProvider.ts
mrjvs, line 234: forgot about it last review, but this will need to also call `updateMediaPlaying(descriptor, state);`

yarn.lock
mrjvs, line 1: this lock file is now filled with npm registries.
please remove the lock file and replace it with the one copied from the dev branch.
also please use yarn in the future to avoid problems like this

mrjvs (changes requested):
tested everything, seems to work great. just those two review comments left and it can be merged directly

zisra:
Done @mrjvs

mrjvs:
> Done @mrjvs

nice, however those import changes you breaks compilation.
please revert them

mrjvs (approved):

PR #228, opened 2023-03-25
Fix for flixHQ provider

Artikronisz:
Since the beginning of the week there was a change on the FlixHQ provider which now returns non-encrypted results. The consumet api was updated https://github.com/consumet/consumet.ts/pull/319 to handle this. 

Because we have no control on the https://c.delusionz.xyz/ endpoint (to update it to the latest version of the api) this means it is no longer reliable and we need to update it to https://api.consumet.org/

Unsure why, but the calls to api.consumet.org using the FlixHQ route will redirect to 9anime (I did not investigate). 
If we update the code to the [TMDB route](https://docs.consumet.org/rest-api/Meta/the-movie-database/get-movie-episode-streaming-links) we now get results as what FlixHQ used to return (from MixDrop and VidCloud).

I will let the maintainers of the project decide if this provider should be renamed.

### DIFF
--- a/src/backend/providers/flixhq.ts
+++ b/src/backend/providers/flixhq.ts
@@ -8,10 +8,7 @@ import {
 } from "../helpers/streams";
 import { MWMediaType } from "../metadata/types";
 
-// const flixHqBase = "https://api.consumet.org/movies/flixhq";
-// *** TEMPORARY FIX - use other instance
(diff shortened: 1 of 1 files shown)


mrjvs:
under the hood the tmdb route still uses flixhq, so I think the naming can stay the same.
code looks good, but haven't had the chance to test it locally.

ill merge it when I get to testing.

mrjvs (approved):

PR #230, opened 2023-03-26
Better subtitle handling

frost768:
- Added a new subtitle parser library
- File types are not limited to srt and vtt
- Fixes malformed subtitle files
- Clean up code
Note: ignore fix safari fullscreen commit, reverted in merge


src/backend/helpers/streams.ts
```diff
@@ -20,7 +20,7 @@ export enum MWStreamQuality {
 export type MWCaption = {
   needsProxy?: boolean;
   url: string;
-  type: MWCaptionType;
+  type?: MWCaptionType;
```
mrjvs, line 23: instead of making it nullable, please just add a "UNKNOWN" enum value.

src/backend/helpers/captions.ts
```diff
@@ -1,52 +1,39 @@
...
-import { MWCaption, MWCaptionType } from "@/backend/helpers/streams";
-import toWebVTT from "srt-webvtt";
+import { MWCaption } from "@/backend/helpers/streams";
 import DOMPurify from "dompurify";
+import { list, parse, detect } from "subsrt-ts";
+import { ContentCaption } from "subsrt-ts/dist/types/handler";
 
+export const subtitleTypeList = list()
+  .map((v) => `.${v}`)
+  .join(",");
```
mrjvs, line 9: perhaps make this an array of file extensions instead of a string. You can join it when you pass it into the accept param of the file upload

src/backend/helpers/captions.ts
```diff
@@ -1,52 +1,39 @@
...
   }
 }
+
+export function parseSubtitles(text: string): ContentCaption[] {
+  if (detect(text) === "") {
+    throw new Error("Invalid subtitle format");
+  }
+  return parse(text)
+    .filter((cue) => cue.type === "caption")
+    .map((cue) => cue as ContentCaption);
```
mrjvs, line 38: instead of mapping (which actually runs code). just cast the entire result of the filter so its just doing things in compilation.

mrjvs (changes requested):
nice job overal, just a few tid bits to fix.

src/backend/helpers/streams.ts
frost768: I was going to remove it but left it. Reason is it is not used anywhere anymore. Also with this new library there aren't any limitations for subtitle file types.

src/video/components/popouts/CaptionSelectionPopout.tsx
```diff
@@ -41,35 +42,17 @@ export function CaptionSelectionPopout(props: {
     async (caption: MWCaption, isLinked: boolean) => {
       const id = makeCaptionId(caption, isLinked);
       loadingId.current = id;
-      controls.setCaption(id, await getCaptionUrl(caption));
+      const blobUrl = await getCaptionUrl(caption);
+      const result = await fetch(blobUrl);
+      const text = await result.text();
+      parseSubtitles(text); // This will throw if the file is invalid
+      controls.setCaption(id, blobUrl);
       controls.closePopout();
```
frost768, line 50: sometimes this doesn't close the popout considering adding a timeout. Should I?

src/backend/helpers/captions.ts
frost768: wouldn't it run map on every render?

mrjvs: I said to only do the joining inside the component. and yes it would. but joining is not expensive to run and realistically it wouldnt render that much at all. basically only when its open.

src/backend/helpers/streams.ts
mrjvs: I would still keep it, even if unused. its valuable information for if we ever switch sub library again or want to preprocess it or something. dont wanna lose the information.

src/video/components/popouts/CaptionSelectionPopout.tsx
mrjvs: yea sure, add a comment explaining why the timeout is tho

mrjvs:
I saw a few review comments resolved but no new commits. is that intentional?

frost768:
> I saw a few review comments resolved but no new commits. is that intentional?

yeah used it as a reminder to myself that i edited them

mrjvs (approved):

PR #231, opened 2023-03-26
Exclude dev routes from production

### LINKED ISSUE #208: [Feature]: Protect dev routes in production

Prevent end users from entering into development pages

Will make it safer and probably will prevent people from exploiting unwanted proxy usage ?

_No response_

frost768:
Resolves #208 
Lazy loading dev componets only for development should decrease chunk size (only by 2 kB).
Removes dev routes from production

### DIFF
--- a/src/setup/App.tsx
+++ b/src/setup/App.tsx
@@ -40,7 +40,7 @@ function App() {
                 />
 
                 {/* other */}
-                {process.env.NODE_ENV === "development" && (
+                {process.env.NODE_ENV === "development" ? (
                   <>
                     <Route
                       exact
@@ -78,7 +78,7 @@ function App() {
                       )}
                     />
                   </>
-                )}
+                ) : null}
                 <Route path="*" component={NotFoundPage} />
               </Switch>
             </Layout>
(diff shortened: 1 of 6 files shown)


src/setup/App.tsx
```diff
@@ -44,15 +40,45 @@ function App() {
...
-                <Route exact path="/dev/test" component={TestView} />
-                <Route exact path="/dev/video" component={VideoTesterView} />
-                <Route
-                  exact
-                  path="/dev/providers"
-                  component={ProviderTesterView}
-                />
-                <Route exact path="/dev/embeds" component={EmbedTesterView} />
+                {process.env.NODE_ENV === "development" && (
+                  <>
```
mrjvs, line 44: please do `process.env.NODE_ENV === "development" ? < /> : null` instead of `process.env.NODE_ENV === "development" && < />`.

right now its actually rendering the text node "false" inside the router if its production (even though you cant see it.

mrjvs (changes requested):
just one tiny change needed

mrjvs (approved):
lgtm

PR #238, opened 2023-03-28
Media session added

### LINKED ISSUE #176: [Feature]: MediaSession implementation

Should add support for navigator.MediaSession.
That way the OS knows how to interact with the media, like playing/pausing through ios control center.

Good ux

_No response_

frost768:
Related to / fix #176 
- Video element poster attribute utilized with backdrop images
- Media session play /pause / seek support
- Changed keys of two div elements which were causing errors


src/video/state/types.ts
```diff
@@ -60,10 +60,13 @@ export type VideoPlayerState = {
...
-      url: string;
-      id: string;
-    };
+    caption:
+      | undefined
+      | null
+      | {
+          url: string;
+          id: string;
+        };
```
mrjvs, line 69: please keep it as only null or a value. lets not do both undefined and null, thats just confusing

src/backend/metadata/justwatch.ts
```diff
@@ -63,6 +65,9 @@ export function formatJWMeta(
         })
       );
   }
+  const backdrops = media.backdrops?.map((u) =>
+    `${JW_BACKDROP_BASE}${u.backdrop_url}`.replace("{profile}", "s1440")
+  );
```
mrjvs, line 70: I think it's better to only have a single nullable backdrop url on the meta object instead of all backdrops.
we're only ever going to use the first one in the list anyway.

src/video/components/internal/VideoElementInternal.tsx
```diff
@@ -41,6 +42,11 @@ function VideoElement(props: Props) {
       autoPlay={props.autoPlay}
       muted={mediaPlaying.volume === 0}
       playsInline
+      poster={
+        meta && meta.meta.meta.backdrops && meta.meta.meta.backdrops.length
+          ? meta.meta.meta.backdrops[0]
+          : ""
+      }
```
mrjvs, line 49: default it to `undefined` instead of empty string. with `undefined` the attribute will actually be unset by react instead of setting it to an empty string.

mrjvs (changes requested):
See review comments, and also I have one question:

Do you think we also need to do the changes for the castingStateProvider?

src/backend/metadata/justwatch.ts
frost768: i thought we would use them but ok

frost768:
> See review comments, and the linting errors.
> 
> Also I have one question: Do you think we also need to do the changes for the castingStateProvider?
> 
> and if we do need almost the exact same code for the chrome casting, maybe it can be moved to a renderless internal video component.

i don't own a casting device so i'm not sure how it works

mrjvs:
> i don't own a casting device so i'm not sure how it works

I think we'd need to do the same to the chrome casting state provider.
But thats best moved into a internal video component. itll just use the state stuff just like the actions to register the media session events. That way we also keep it modular, easy to change things in the future.

mrjvs:
As this PR has requested changes and haven't been modified for more than a week. I'm gonna close it, feel free to reopen or make a new PR when the changes have been made.

frost768:
I did change the code with suggested ones accordingly. Are you talking about moving it to a different component that is wrapped in videoplayerbase component?

mrjvs:
> I did change the code with suggested ones accordingly. Are you talking about moving it to a different component that is wrapped in videoplayerbase component?

yea, perhaps I wasn't clear enough that I think that should be done that way.
We should move the code from the stateProviders to an internal video component put in the base

frost768:
Opened a new one with requested changes #254

PR #239, opened 2023-03-28
Remove isFocused verification for Keyboard shortcuts

judemont:
I removed the isFocused verification when using a keyboard shortcut because it didn't allow to do a keyboard shortcut when the video is not focused.


JipFr:
We want to allow for multiple video elements in a page. The video should be focused for keyboard shortcuts.

judemont:
Okay, I understand.
It should be that by default if there is no video focus the keyboard shortcuts work for all videos.

JipFr:
By your logic, pressing the spacebar on page load would play a dozen videos at once. That doesn't seem right to me, keyboard shortcuts should only activate a single element.

judemont:
That's right.

PR #243, opened 2023-03-30
Add volume up / down keyboard events

JipFr:


mrjvs:
idk what happened but you have 9k changes to the lock file. and its causing the linting to fail on unchanged files.
can you revert the lock file changes?

mrjvs (approved):

Issue #237, opened 2023-03-27
[Bug]: No captions in picture in picture


JipFr:
### What happened?

Because we have custom subs, no subtitles show up in picture of picture.

### What browsers are you seeing the problem on?

Chrome

### Steps to reproduce?

1) Pick subtitles
2) Open Picture in Picture

frost768:
Does chrome support subtitles in PiP mode? I have only seen it on Firefox

JipFr:
I thought it did, but I can't seem to replicate it at the moment. Still, even if it is only supported for Firefox (and Safari), we should still implement it

frost768:
I did some googling and found that firefox doesn't support `document.pictureInPictureElement` property. I was going to use it to conditionally add track element to video. They use `MozTogglePictureInPicture` event on video element but it only fires when getting into PiP mode not exiting it. So if somehow we can detect the event of exiting PiP mode we can do it.

frost768:
Seems like they won't support standart PiP spec with their own "user-oriented" PiP implementation 
https://bugzilla.mozilla.org/show_bug.cgi?id=1611831

PR #250, opened 2023-04-03
Subtitle tests and type controls

### LINKED ISSUE #269: [Bug]: First line of multiline captions are ignored by the library
### What happened?

Subtitles with multiple lines are not parsed correctly. The first line is always ignored. This only occurs on files with LF line ending.
When converted to CRLF it works fine
Example:
Input
```
WEBVTT

00:00:00.000 --> 00:00:10.000
- Test 1\n- Test 2\n- Test 3

00:00:10.000 --> 00:00:20.000
- Test 1
- Test 2
- Test 3
```
Output:
```json
[{
    "type": "caption",
    "index": 1,
    "start": 0,
    "end": 10000,
    "duration": 10000,
    "content": "- Test 2 - Test 3",
    "text": "- Test 2 - Test 3"
 },
{
    "type": "caption",
    "index": 2,
    "start": 10000,
    "end": 20000,
    "duration": 10000,
    "content": "- Test 2 - Test 3",
    "text": "- Test 2 - Test 3"
 }]
```

Firefox, Chrome, Safari, Microsoft Edge

### Steps to reproduce?

_No response_

~I have opened an issue in the library repo hopefully they'll look into it
https://github.com/leranjun/subsrt-ts/issues/2~
They looked into it

frost768:
- Filter non-subtitle files in providers
- Reset caption delay on every subtitle change and exiting video player. Delay should not persist between video sessions
- Add tests for subtitles
- ~Use forked version of subsrt-ts with the multiline support patch~ subsrt-ts updated (fix #269 )


mrjvs:
looks good, however I don't see how this will fix #237.
It's still missing a track element?

frost768:
> looks good, however I don't see how this will fix #237. It's still missing a track element?
it doesnt sorry, we can do it using pip events but it would be incompatible with firefox. there is no way to detect leaving the pip mode only entering it

TheGloriousDuck:
> > looks good, however I don't see how this will fix #237. It's still missing a track element?
> 
> it doesnt sorry, we can do it using pip events but it will be incompatible with firefox. there is no way to detect leaving the pip mode on firefox only entering it

I'm looking here and it looks like it does?? I don't know, I could be wrong
https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API#events

TheGloriousDuck:
I see. Could we still just implement it for browsers other than Firefox then? Better to have something than nothing

frost768:
How can we clean the node_modules cache in the pipeline? It compares the official package with the forked one.

src/backend/providers/superstream/index.ts
```diff
@@ -111,6 +115,30 @@ const getBestQuality = (list: any[]) => {
   );
 };
 
+const convertSubtitles = (subtitleGroup: any): MWCaption | null => {
+  let subtitles = subtitleGroup.subtitles;
+  subtitles = subtitles
+    .map((subFile: any) => {
```
JipFr, line 121: Do we require explicit `any` / should these be typed?

src/backend/providers/superstream/index.ts
```diff
@@ -111,6 +115,30 @@ const getBestQuality = (list: any[]) => {
...
+      if (!supported) return null;
+      const type = getMWCaptionTypeFromUrl(subFile.file_path);
+      return {
+        ...subFile,
+        type: type as MWCaptionType,
+      };
+    })
+    .filter(Boolean);
+
+  if (!subtitles.length) return null;
```
JipFr, line 132: I feel like using a falsy check here feels a little weird, perhaps the following instead?
```suggestion
  if (subtitles.length === 0) return null;
```

src/utils/thumbnailCreator.ts
```diff
@@ -0,0 +1,41 @@
...
+  const video = document.createElement("video");
+  video.src = videoUrl;
+  video.crossOrigin = "anonymous";
+
+  // Wait for the video metadata to load
+  const metadata = await new Promise((resolve, reject) => {
+    video.addEventListener("loadedmetadata", resolve);
+    video.addEventListener("error", reject);
+  });
+  console.log(metadata);
```
JipFr, line 14: Console.log should probably be removed
```suggestion
```

src/video/components/actions/CaptionRendererAction.tsx
```diff
@@ -63,20 +63,37 @@ export function CaptionRendererAction({
...
     } else {
       captions.current = [];
     }
   }, [source?.caption?.url]);
-
+  useEffect(() => {
+    // reset delay after video ends
+    return () => setCaptionDelay(0);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
JipFr, line 76: Why is exhaustive deps disabled? What iss the problem with adding `setCaptionDelay` as a dep?

frost768: it causes too much rendering. not sure why

mrjvs: I wouldn't make a unmount to reset, I'd rather set it to 0 on video player load and when source url changes.

src/utils/thumbnailCreator.ts
frost768: i think the whole file should be removed. it doesn't have anything to with subtitles and is not complete as a feature

src/video/components/actions/CaptionRendererAction.tsx
frost768: So should I put the useEffect it in `VideoPlayer.tsx`?

src/backend/providers/superstream/index.ts
frost768: As @mrjvs said in discord the providers are prone to change so I think it should stay as any

src/utils/thumbnailCreator.ts
JipFr: Jvs and I agree — remove it. We’re excited to see it in the app though!

src/video/components/actions/CaptionRendererAction.tsx
cruizm-cyber: hey ladys

mrjvs (approved):

Issue #163, opened 2023-02-21
[Feature]: A settings page


mrjvs:
### What feature do you want to add?

A settings page, showing these items:
 - [x] Current version the app is running
 - [x] A language selector
 

### Why do you want to have this feature?

there is currently no way to see/edit the aforementioned items

maxwellward:
Rather than a dedicated page, if it's just those two options could we put them in the footer instead?

mrjvs:
> Rather than a dedicated page, if it's just those two options could we put them in the footer instead?

I was thinking a full page or modal, so we don't have to keep finding places to put settings when we get new ones. A bit of future proofing.

athaller:
Was thinking about having a setting for captions on or off by default and glad I came across this topic.

Other options that can be added:
 - Autoplay On/Off Toggle
 - Captions On/Off (Default option)
 - Default/Recommended Provider
 - Default/preferred quality (as I think right now everything is hardcoded for 1080p from a quick src glance)
 - Option to skip intro/credits of tv shows (but would require a feature ticket to have the base functionality added)

judemont:
A settings page is a good idea,
here are some other elements that could be added in the settings:
- Change keyboard shortcuts
- Disable/enable history
- Choose preferred providers

Issue #177, opened 2023-02-27
[Feature]: Subtitle preference saving


mrjvs:
### What feature do you want to add?

Save language for subtitles and auto apply when loading in

### Why do you want to have this feature?

Good ux, recommendation of a coworker

PR #252, opened 2023-04-06
A settings modal

frost768:
This PR is for the the following issues
 #163 and #177 

- Settings modal
   - Language selection
   - Default subtitle language selection
   - Subtitle style change with a preview display

-  Fetch default subtitle with a retry mechanism


frost768:
Desktop and mobile preview. Responsive in-between
desktop
mobile

src/video/components/actions/CaptionRendererAction.tsx
```diff
@@ -22,9 +22,14 @@ function CaptionCue({ text }: { text?: string }) {
...
     <p
-      className="pointer-events-none mb-1 select-none rounded px-4 py-1 text-center [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]"
+      className={[
+        "pointer-events-none mb-1 select-none rounded px-4 py-1 text-center [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]",
+      ].join(" ")}
       style={{
         ...captionSettings.style,
+        fontSize: !scale
+          ? captionSettings.style.fontSize
+          : captionSettings.style.fontSize * scale,
```
mrjvs, line 32: I prefer doing it like this: `captionSettings.style.fontSize * ( scale ?? 1)`. Less conditional branches can be easier to read

src/views/SettingsModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+    setCaptionBackgroundColor,
+    setCaptionColor,
+    setCaptionFontSize,
+  } = useSettings();
+  const { t, i18n } = useTranslation();
+
+  const colors = ["#ffffff", "#00ffff", "#ffff00"];
+  const selectedCaptionLanguage = useMemo(
+    () => captionLanguages.find((l) => l.id === captionSettings.language)!,
+    [captionSettings.language]
```
mrjvs, line 30: Please dont use non-null assertions

mrjvs: same file at the dropdown there is another one

src/views/SettingsModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+                          className={[
+                            "absolute text-xs text-[#1C161B]",
+                            color === captionSettings.style.color
+                              ? ""
+                              : "hidden",
+                          ].join(" ")}
+                          icon={Icons.CHECKMARK}
+                        />
+                      </div>
+                    ))}
```
mrjvs, line 132: As the color swatch button is no longer used in only one place, I think its justified to make that a standalone component.

src/views/SettingsModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+            <div className="flex w-full flex-col justify-center">
+              <div className="flex aspect-video flex-col justify-end rounded bg-zinc-800">
+                {selectedCaptionLanguage.id !== "none" ? (
+                  <div className="pointer-events-none flex w-full flex-col items-center transition-[bottom]">
+                    <CaptionCue
+                      scale={0.5}
+                      text={selectedCaptionLanguage.nativeName}
+                    />
+                  </div>
+                ) : null}
```
mrjvs, line 147: I think that we should still show a caption cue preview even if the preferred caption language is unset. just use `lorem ipsum` as some placeholder text if the language isnt set

src/views/SettingsModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+    language,
+    setLanguage,
+    setCaptionLanguage,
+    setCaptionBackgroundColor,
+    setCaptionColor,
+    setCaptionFontSize,
+  } = useSettings();
+  const { t, i18n } = useTranslation();
+
+  const colors = ["#ffffff", "#00ffff", "#ffff00"];
```
mrjvs, line 27: this is now used in two places, please put it in a file somewhere so we don't have the same constant color definitions is two places

src/views/SettingsModal.tsx
mrjvs, line 1: Can you add the version number of the app to the settings modal somewhere, that way we fully resolve #163.
You can use the `conf()` function to get it.

mrjvs (changes requested):
looks good, just a few requests for code cleaniness.

I do think this needs a redesign in the future, a simple modal probably won't be enough if it grows.
But as it is right now, its plenty for a first version.

src/views/SettingsModal.tsx
frost768: i think we should not because it represents how it will look like in video player. it sounds silly but some may think that it will show a text even if a caption is not selected

src/views/SettingsModal.tsx
frost768: should we also move colors array into this component? it will only be used in this.

src/video/components/actions/CaptionRendererAction.tsx
```diff
@@ -22,9 +22,14 @@ function CaptionCue({ text }: { text?: string }) {
 
   return (
     <p
-      className="pointer-events-none mb-1 select-none rounded px-4 py-1 text-center [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]"
+      className={[
+        "pointer-events-none mb-1 select-none rounded px-4 py-1 text-center [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]",
+      ].join(" ")}
```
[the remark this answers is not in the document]
frost768: i was trying to use classes for scaling but you are right it is not necessary anymore

src/views/SettingsModal.tsx
mrjvs: I wouldn't, however you can put the constant export into the same file as the component

frost768: what about other caption settings? they all use slider component should i also extract them to their own component?

zisra:
Would it make sense to add playback speed?

frost768:
> Would it make sense to add playback speed?

without a dummy video, i dont think so

src/views/SettingsModal.tsx
mrjvs: I don't think we should make the block of sliders a component. just do the colorSwatch button and constant export for the color list.

mrjvs:
> what would be nice as well is to change the homepage dropdown default, I personally don't really watch movies, so I think being able to set series as default for the search bar would be nice

Please make that a new issue, its not in scope for this PR

mrjvs (changes requested):
looks good, just change the denim-200 to denim-300 for the background and then its all done

mrjvs (approved):

PR #254, opened 2023-04-09
Media session component

frost768:
Subsequent to PR #238
Moved media session logic into its own component


src/video/components/internal/MediaSessionInternal.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return () => {
+      mediaSession.metadata = null;
+      mediaSession.playbackState = "none";
+      mediaSession.setActionHandler("play", null);
+      mediaSession.setActionHandler("pause", null);
+      mediaSession.setActionHandler("seekbackward", null);
+      mediaSession.setActionHandler("seekforward", null);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [state.stateProvider]);
```
mrjvs, line 78: this will break at changes in meta data or similar.
What you can do is make a media session once with this component (0 dependencies). put it in a ref.
then use a bunch of useEffects to update the metadata in the session or update a few refs.

You can use refs freely in effects without adding them to the dependency array.

So in short:
 - one effect to make the session, store it in a ref
 - useEffect on meta data to update the media session ref
 - useEffect on useControls and useProgress and store both in refs
 - the media session handlers can use the progress and controls refs to control things

mrjvs (changes requested):
just one change to prevent it breaking unexpectedly

src/video/components/internal/MediaSessionInternal.tsx
frost768: meta updates with progress. if i use meta as a dependency it will run every second causing media session metadata to change along with it.

mrjvs: you can memoize the only the metadata you use and use that as a dependency

frost768: yeah you can but the memoized version will be watching the original one for changes hence it will trigger no matter what

mrjvs: its a bit jank, but if you stringify before memoing, it will only trigger on change `useMemo(() => JSON.stringify({ postUrl: "" }))`

mrjvs:
Pull request is out of date with v4. Closing

(sorry that I never got it again)

PR #262, opened 2023-04-15
Add serverf4 provider

zisra:
**This should be working (tested with no-CORS extension) but there is one issue:**
- The provider needs to make a request to an URL, not to read the contents, but to see what URL it finally redirects to. 
- Usually when requesting a URL with redirects, the fetch api would return an object that contains the final URL including redirects. 
- But when using a CORS proxy, redirects are handled by the proxy and the final URL is the same as the URL requested

**Possible solutions**
- Don't handle redirects with the proxy and make the client handle them
- Add a `X-Final-Destination` header to the response

### DIFF
--- a/src/backend/helpers/fetch.ts
+++ b/src/backend/helpers/fetch.ts
@@ -29,7 +29,11 @@ export function mwFetch<T>(url: string, ops: P<T>[1] = {}): R<T> {
   return baseFetch<T>(url, ops);
 }
 
-export function proxiedFetch<T>(url: string, ops: P<T>[1] = {}): R<T> {
+export function proxiedFetch<T>(
+  url: string,
+  ops: P<T>[1] = {},
+  raw = false
+): R<T> {
   let combinedUrl = ops?.baseURL ?? "";
   if (
     combinedUrl.length > 0 &&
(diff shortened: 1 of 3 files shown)


mrjvs:
Nice work, good find.

For the proxy issues, I think it's best we make an extra header `X-Disable-Redirect-Following: "true"`.
we have a repository for the proxy here: https://github.com/movie-web/simple-proxy, if you feel up for it, you can PR it there. should be a one-liner.

src/backend/helpers/fetch.ts
```diff
@@ -29,7 +29,11 @@ export function mwFetch<T>(url: string, ops: P<T>[1] = {}): R<T> {
   return baseFetch<T>(url, ops);
 }
 
-export function proxiedFetch<T>(url: string, ops: P<T>[1] = {}): R<T> {
+export function proxiedFetch<T>(
+  url: string,
+  ops: P<T>[1] = {},
+  raw = false
+): R<T> {
```
mrjvs, line 36: I think it's better to have a `rawProxiedFetch(url, ops): FetchResponse<any>` instead of modifying the original one with a boolean option.

mrjvs: That would also fix the type errors

src/backend/providers/serverf4.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const videoInfo = await proxiedFetch<any>(
+      `https://serverf4.org/api/source/${id}`,
+      {
+        method: "POST",
+      }
+    );
+
+    const source = videoInfo.data[videoInfo.data.length - 1];
+
```
mrjvs, line 61: make a specific check if there are no files and throw a error, much like the no iframe error. that way its easier to trace what the problem is when it fails

mrjvs (changes requested):

src/backend/helpers/fetch.ts
zisra: I'm having trouble with this and can't figure out how to fix this redirect thing, sorry

zisra:
Not sure what happened to serverf4.org but it's gone

PR #268, opened 2023-04-20
V3.0.11

mrjvs:
- add hdwatched provider
- adjust volume with keybinds now has an extra popup to show the percentange
- timestamp click to change rendering
- fix idle mouse bug
- settings modal


JipFr (approved):
Beautiful!

PR #274, opened 2023-04-23
language based time formatting

frost768:
Instead of using fixed 12 hour time format, use language based time format


JamesHawkinss (changes requested):
Thanks for this PR. I just merged some lang PRs so there's a few new language files on dev which don't support this new formatting, could you update your PR accordingly? Happy to merge once that is done.

PR #279, opened 2023-04-25
fix: Chrome cast handling, useEffect cleanup, and key prop issues

yilmazcabuk:
### Description

This pull request addresses several issues and makes improvements to the codebase:

- Refactors the Chrome cast handling in `CastingInternal.tsx` for improved robustness and error prevention
- Adds cleanup functions to `useEffect` hooks in `useScrape`, `BackdropAction`, and `HomeView` to prevent memory leaks
- Fixes `key` prop issues in `CaptionSelectionPopout`, `CaptionSettingsPopout`, and `SettingsModal` components for rendering consistency

### Changes Made

- Updated `CastingInternal.tsx` to wrap casting logic with conditional check for `chrome.cast` and `chrome.cast.isAvailable`
- Added cleanup functions to `useEffect` hooks in `useScrape`, `BackdropAction`, and `HomeView` to clear timeouts on unmount and dependency change
- Fixed `key` prop issues in `CaptionSelectionPopout`, `CaptionSettingsPopout`, and `SettingsModal` components by updating `key` prop values for uniqueness and consistency


src/video/components/actions/BackdropAction.tsx
```diff
@@ -90,6 +90,9 @@ export function BackdropAction(props: BackdropActionProps) {
       lastBackdropValue.current = currentValue;
       props.onBackdropChange?.(currentValue);
     }
+    return () => {
+      if (timeout.current) clearTimeout(timeout.current);
+    };
```
mrjvs, line 95: has this change been tested, does idle mouse movement still work?
cuz afaik, this will make the timeout always clear.

src/video/components/internal/CastingInternal.tsx
```diff
@@ -36,34 +36,42 @@ export function CastingInternal() {
     const state = getPlayerState(descriptor);
     if (!available) return;
 
-    state.casting.instance = cast.framework.CastContext.getInstance();
-    state.casting.instance.setOptions({
-      receiverApplicationId: chrome.cast.media.DEFAULT_MEDIA_RECEIVER_APP_ID,
-      autoJoinPolicy: chrome.cast.AutoJoinPolicy.ORIGIN_SCOPED,
-    });
+    if (
```
mrjvs, line 39: does this actually help anything, we already have an isAvailable hook to check this, which it early returns on L37.

src/hooks/useScrape.ts
```diff
@@ -39,32 +40,42 @@ export function useScrape(meta: DetailedMeta, selected: SelectedMediaData) {
...
+            setEventLog((arr) => [
+              ...arr,
+              {
+                errored: false,
+                id: ctx.id,
+                eventId: ctx.eventId,
+                type: ctx.type,
+                percentage: 0,
+              },
+            ]);
```
mrjvs, line 53: please use the guard clause pattern in this file to prevent nesting as much as possible.
so do
```js
if (!condition) return;
...
```

instead of
```js
if (condition) {
  ...
}
```

src/video/components/internal/CastingInternal.tsx
yilmazcabuk: Before this check, streams on mobile view were crushing before being started 'cuz chrome cast was returning undefined. After this patch it solved. But we really need to re-structure project in many points imo. I'm volunteer to do that.

mrjvs: while I appreciate the gesture, a rewrite/restructure of everything is a drastic move. one that I wouldn't do unless there was a very good reason for it. Before just executing it, could make an issue for it elaborating why and what you want to restructure? Then a discussion can be had beforehand.

mrjvs:
why was the repo deleted?

src/video/components/internal/CastingInternal.tsx
mrjvs: also to come back to the chromecasting errors, it seems we rely on the chromecast sdk to let us know if this browser supports it and seems its unreliable, I think the check you added is better placed inside the isAvailable hook

PR #290, opened 2023-05-04
Add "Save watch history" option in the settings

judemont:
Hello, 
I have added an option in the settings to choose if you want to save the viewing history or not.

I find this option useful because it allows the user to be able to choose to not save their history.


JipFr:
Who is this for?

judemont:
> I don't know why anyone would want such a setting.

To ask the user to disable cookies to have no history, poses several problems:
- Some people don't know how to disable cookies (or don't know that it is possible) 
- Some users don't know that the history is saved with a cookie.
- Users can't know if cookies are not used for other things.
- Even if the user knows all this, it is just more pleasant.

This kind of option has in my opinion a place in the settings.

thehairy:
And why would you even want to disable this at all
what benefits do you get from that

judemont:
> Who is this for?



> Who is this for?

There is no particular issue that asks for that, but I think it is a basic option that should be included.

JipFr:
Why though. Are you watching smut on movie-web?

thehairy:
> > Who is this for?
> 
> There is no particular issue that asks for that, but I think it is a basic option that should be included.

Yes, but the question is why.
I don't see a point in this setting besides making your life harder when watching series and movies.
You can't jump back in when your connection gets interrupted or you take a break.
You have no recollection of what you watched so far, except your brain.

And even if, I don't think settings on the video player is the proper place for this.

JipFr:
> And even if, I don't think settings on the video player is the proper place for this.

While I agree the PR seems to be without purpose, it's not actually in the video player. It's in the global settings.

thehairy:
oh, TIL there is a global settings page

judemont:
Okay, no problem👍👍

Issue #236, opened 2023-03-27
[Feature]: Add 2embed.to provider


atozmovies:
### What feature do you want to add?

Add 2embed.to as provider

### Why do you want to have this feature?

it has lots of embeds to choose from, lots of videos.

### Anything other details to share?

it will require scraping from embeds, which can be difficult

lem6ns:
planned

mrjvs:
2embed has not been in the planning, I will rename this issue to match it

frost768:
2embed's vidcloud server videos are restricted to an origin (rabbitstream) with cors header

zisra:
2embed uses recaptcha too

PR #292, opened 2023-05-09
Add 2embed provider

Jordaar:
Added https://2embed.to provider by scraping [Streamlare](https://streamlare.com) links.

Closes: #236 

# Things noticed about 2embed

- The vidcloud provider can’t be implemented because the m3u8 streams are blocked by CORS.
- The quality of Streamlare links varies. Newer movies have cam quality, but older ones have better quality and multiple subtitle options.

# Update required

~~The proxy needs to be updated, refer PR: movie-web/simple-proxy#9~~


mrjvs:
Hey, code looks good.

however after trying the provider out. I can't seem to find any media that work with it, do you have a few examples I can try with?

also, could you explain a bit about how the recaptcha bypass works? it seems to just get the token, but wouldnt that only work if the individual has already been on the site before?

Jordaar:
It appears that 2Embed provides Streamlare links for a limited selection of movies, primarily consisting of older titles such as The Matrix. While I leave the final decision to you, I personally believe that adding this provider would not be beneficial at this point in time.

As for the recaptcha bypass, I don’t have much information, but fetching tokens seems to do the trick here.

Demolliderish:
Hi there!
Just dropping by and saying that's a neat way of bypassing the recaptcha from 2embed.
I'm a bit new to the scraping scene. Curiosity got a hold of more and I tried recreating your project on python.
It was smooth sailing up to the point of getting hold of;
  'followStream.headers.get("X-Final-Destination");' line 185 src/backend/providers/2embed.ts

Interestingly the resulting data for your;
  const followStream = ... (line 180)
  
  returned a JSON which had a promising value called 'location'. In it lies a link for what seemed to be the download call.
  However clicking the link instead sent me to a page with the following error;
              410 Gone
            nginx/1.21.6

The 
  Below is a snippet  of my python code using;
  https://ninjashare.to/download/qFHvfE9b67mPFUE431j9W2?t=51eeb022c0bff1b57c80534020a2518c
  as a temporary starting point 
  
image

zisra:
@Demolliderish the `X-Final-Destination` is actually created by our[ Cloudflare workers proxy](https://github.com/movie-web/simple-proxy/blob/f2d4e523a37761bd4e83dbf7a568f6b6b35fb795/src/main.js#L105). It represents the URL that the request redirects to. You can see a more detailed explanation in this PR I made: #262. 
> **This should be working (tested with no-CORS extension) but there is one issue:**
> 
> * The provider needs to make a request to an URL, not to read the contents, but to see what URL it finally redirects to.
> * Usually when requesting a URL with redirects, the fetch api would return an object that contains the final URL including redirects.
> * But when using a CORS proxy, redirects are handled by the proxy and the final URL is the same as the URL requested
> 
> **Possible solutions**
> 
> * Don't handle redirects with the proxy and make the client handle them
> * Add a `X-Final-Destination` header to the response

I don't know Python, but a working solution would be simply getting the [location header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) from your response

Jordaar:
> Hi there! Just dropping by and saying that's a neat way of bypassing the recaptcha from 2embed. I'm a bit new to the scraping scene. Curiosity got a hold of more and I tried recreating your project on python. It was smooth sailing up to the point of getting hold of; 'followStream.headers.get("X-Final-Destination");' line 185 src/backend/providers/2embed.ts
> 
> Interestingly the resulting data for your; const followStream = ... (line 180)
> 
> returned a JSON which had a promising value called 'location'. In it lies a link for what seemed to be the download call. However clicking the link instead sent me to a page with the following error; 410 Gone nginx/1.21.6
> 
> Below is a snippet of my python code using; https://ninjashare.to/download/qFHvfE9b67mPFUE431j9W2?t=51eeb022c0bff1b57c80534020a2518c as a temporary starting point
> 
> image


From what I can understand, the post request should return a JSON like this:
```json
{
  "status": "success",
  "message": "OK",
  "token": "c959aca4281cac1f32b93753855132a4",
  "result": {
    "Original": {
      "label": "Original",
      "size": 2298049645,
      "url": "https:\/\/larecontent.com\/download?token=SBFGQV8RCRFbR0dDQAlvHG8cREREHgcCRwQLAQYEHUBAXwNXHVBcXm8cYlV2QlVrXVhBfQoeYEZ_AHhWckFLVG8cAgULBwsDAgQGA28cAVIDBQkKC1ADbxx6BnZrZFhpXllmSn5JaXQeR0l1e11DQ0dsBWNrdF1KfHJlWgBHV1BHUmECSVhGYgECfmZpA2FRS2dcUUYFa1l0SVxLd11ASUZGd2BEVFBndVJYSgUDWGQKC0FRflJ1ZXZFfwN4ewUFUnR8RANpQndbZmdECwdQQmprWngLVWBWB3lcZx5DZwdBRQEBXEVJYwpfAGBEVEtQZGoBVH1fcVViY1VsbHQLZHYEaXQEZENWVGZ5cEZnBnVaY2V5VlBScGRRVEcef2tFdl5GHmJBA10GCwUDVR4BfmoAXXhEWkRhX2JHfkJLYgBxXHx0fVBKCgcAAwF7BHx4W1tmf1p0cFVQQB5dfGtXf397B2p4A0t_HlB5QFBxYm8cBQIBCwsLBB1eQwcMV18OAhEfEVpDEQkRAVIDBQkKC1ADEU4"
    }
  }
}
```

The "url" here redirects to an absolute .mp4 URL. To obtain the final .mp4 URL, you can check what the URL redirects to. In this case, it will be:
```
https://www-41t78257.ssl0d.com/YJJgdoe8TvbgQ-71OgYIww/1684803885/2a06:98c0/I5EXWkZmjUyMzZG-tzFHnppt_6PXGnyOAVi3tdctaR1zkuQ21MUZ0RbxTobu6XjGzoxDnszuuDSwgcTFaky60kW98rbMaFVEvL0KH66aGOw0ZqDhUTw84cqYXiK8fSe4JoT-pT4rv22ovzP9l3SwgxcWY2gNlBfQPf__G8WE7ZG7WpegUJCuT5FiPVJecaCWbgt-LXvEmu-Qr0n5860f-2MY3nKwiwRlQtMqxQ3BoOGNcy94302H7OKhhULiGCfcs-nOXdLLH4YK0xL-cJscBQ/6128887.mp4?dl=1
```

Please note that these links are only valid for one-time use and will expire afterward, returning a "410 Gone" status.

mrjvs (approved):

Demolliderish:
**'Please note that these links are only valid for one-time use and will expire afterward, returning a "410 Gone" status.'**

That is problem, despite being my first time opening the link - it returns me a '410 gone' status.
Any ideas what this occurs?

erynith:
> 

but doesn't MW use a cors proxy so how is the stream still blocked by cors 🥲

PR #300, opened 2023-05-21
fix: move meta id check to providers

thehairy:
This PR removes the tmdbId from DetailedMeta and moves the imdbId check to the providers that require them.

*This PR has been tested and no errors have been found. Typings have been properly changed.*

### DIFF
--- a/src/backend/metadata/getmeta.ts
+++ b/src/backend/metadata/getmeta.ts
@@ -29,8 +29,7 @@ interface JWDetailedMeta extends JWMediaResult {
 
 export interface DetailedMeta {
   meta: MWMediaMeta;
-  tmdbId: string;
-  imdbId: string;
+  imdbId?: string;
 }
 
 export async function getMetaFromId(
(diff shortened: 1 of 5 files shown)


src/backend/metadata/getmeta.ts
```diff
@@ -29,8 +29,7 @@ interface JWDetailedMeta extends JWMediaResult {
 
 export interface DetailedMeta {
   meta: MWMediaMeta;
-  tmdbId: string;
-  imdbId: string;
+  imdbId?: string;
 }
```
mrjvs, line 33: even though we aren't using tmdbId right now, I think we should keep it in as an optional field, just in case some provider does rely on it in the future.

We wanna keep as many options for providers open as possible.

mrjvs (changes requested):

src/backend/metadata/getmeta.ts
thehairy: I have re-added tmdbId for potential future use.

mrjvs (approved):

PR #304, opened 2023-05-22
Drop subtitle file

zisra:
Screen Recording 2023-05-22 at 6 06 45 PM

- **Allow dropping subtitle files, for better UX. This makes it very easier for me**
- Pressing `esc` closes all popouts. This is pretty much standard


thehairy (approved):
Looks good, but you might as well complete the other translations too, deepl should be quite accurate for this.

kach17:
Can there be some hint in UI that drag and drop is available?

zisra:
I would love to, just can't find  a way to

JamesHawkinss:
> I would love to, just can't find a way to

@mrjvs ux needed

binaryoverload:
Thanks for your PR, but as there isn't an issue for this, we will close this for now. 

If an issue is accepted for this feature, you can ask to be assigned to it.

PR #305, opened 2023-05-24
Add Sflix provider

Jordaar:
Added [sflix.video](https://sflix.video) provider.

- Newer movies are added earlier than other providers.
- Most of the movies are less than one year old, and the selection is limited.


thehairy (dismissed):
Looking good :)

mrjvs:
works great, but their servers are very slow, can we put it at the bottom of the ranking so it only gets tried as a last priority?

Jordaar:
Okay

mrjvs (approved):

PR #306, opened 2023-05-25
Add T query param for time and make scrollbar styles global

JipFr:
- Add support for a `t` query param, allowing people to share links with a specific time. Supports both seconds (like how YouTube does it, say `t=530` and MM:SS and HH:MM:SS, e.g `?t=1:20:05`
- Make scrollbar styles global


src/video/components/controllers/ProgressListenerController.tsx
```diff
@@ -56,9 +58,28 @@ export function ProgressListenerController(props: Props) {
...
+    if ((queryParams.t ?? null) !== null) {
+      // Convert `t` param to time. Supports only seconds, but also `3:30` or `1:30:02`
+      const timeArr = queryParams.t.toString().split(":").map(Number);
+
+      const hours = timeArr[timeArr.length - 3] ?? 0;
+      const minutes = Math.min(timeArr[timeArr.length - 2] ?? 0, 59);
+      const seconds = Math.min(
+        timeArr[timeArr.length - 1] ?? 0,
+        minutes > 0 ? 59 : Infinity
+      );
```
binaryoverload, line 73: Jank

binaryoverload (changes requested):

src/video/components/controllers/ProgressListenerController.tsx
binaryoverload: If you reverse the array instead, then you will be able to use natural indexing

binaryoverload (dismissed):
I can now wholeheartedly endorse this PR

src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,22 @@
...
+  const queryParams = useMemo(() => {
+    // Basic absolutely-not-fool-proof URL query param parser
+    const obj: Record<string, string | number> = {};
+    for (const [key, value] of loc.search
+      .slice(1)
+      .split("&")
+      .map((e) => e.split("="))) {
+      const valueAsNum = Number(value);
+      obj[key] = Number.isNaN(valueAsNum) ? value : valueAsNum;
+    }
```
mrjvs, line 16: any reason why you dont just `return new UrlSearchParams(loc.search)`?

JipFr: Only because I did not know it was a thing :P

mrjvs (approved):

Issue #295, opened 2023-05-12
[Bug]: subtitle parts that have same key dont clear


RihanArfan:
### What happened?

After malformed/multiple subtitles show at once, they're not cleared.

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge

### Steps to reproduce?

1. Series
2. Mob Psycho 100 (https://movie-web.app/media/JW-show-39013/98611/1636826)
3. Season 2
4. Set English subtitles
5. See 1m:35s - 1m42s for a large number of subtitles (expected) which don't clear (bug)

### Other relevant information

Viewing the media directly on FlixHQ, the subtitles disappear when the scene changes

zisra:
<img width="1458" alt="image" src="https://github.com/movie-web/movie-web/assets/100528712/2af3ecf8-a4a2-4085-a933-bb0338b3645b">
🤔

frost768:
```
34
0:01:40,40 --> 0:01:43,61
Executing the Exorcism!!

35
0:01:40,40 --> 0:01:43,61
Refer us to a friend now and get an extra 5% reduction!! 
Also get an add-on exorcism for 20% off!!

36
0:01:40,40 --> 0:01:43,61
We will do our best to reduce the spirit possessing you
Option A — Trial Course\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\hA Mere 2,000 yen (Tax Excluded)
\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\hOutside Offer Period: 3,980 yen
Option B — Serious Course \h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\h\hA Mere 5,000 yen (Tax Excluded)
[...]

PR #308, opened 2023-05-26
Subtitle ad-blocking

zisra:
Test:
- /media/JW-movie-75152 
- Go to 8 seconds
- Usually there would be an opensubtitles link here


JamesHawkinss:
@mrjvs ops on filtering ads / credits? I'm not sure if it's the right thing to do

zisra:
Should fix #295 if that helps

mrjvs:
I don't think it's ethical to block subtitle credits and ads.
The subtitle makers deserve some credit.

Thanks for the PR, but im gonna close it.

PR #321, opened 2023-06-06
External subtitles support

zisra:
I'm stuck on loading subtitles


mrjvs:
I think we should scrape subscene ourselves instead of using justchills API (which just calls subscene)
They are already struggling with server costs, it'd be rude to scrape them for subtitles.

PR #322, opened 2023-06-08
Thumbnails

### LINKED ISSUE #223: [Feature]: add precise seeking with thumbnails

precisee seeking with thumbnails

its a indispensable feature

_No response_

frost768:
resolves #223 
Preview:

https://github.com/movie-web/movie-web/assets/19949367/9e607bf7-970f-4962-aae8-46f2e1032818


src/utils/thumbnailCreator.ts
```diff
@@ -0,0 +1,51 @@
...
+): AsyncGenerator<Thumbnail, Thumbnail> {
+  const canvas = document.createElement("canvas");
+  const ctx = canvas.getContext("2d");
+  if (!ctx) return { from: -1, to: -1, imgUrl: "" };
+  const video = document.createElement("video");
+  video.src = videoUrl;
+  video.crossOrigin = "anonymous";
+
+  // Wait for the video metadata to load
+  const metadata = await new Promise((resolve, reject) => {
```
thehairy, line 19:
```suggestion
  await new Promise((resolve, reject) => {
```

metadata is not used, linter is complaining.

thehairy (changes requested):

thehairy (approved):

src/utils/thumbnailCreator.ts
```diff
@@ -0,0 +1,51 @@
+export interface Thumbnail {
+  from: number;
+  to: number;
+  imgUrl: string;
+}
+export const SCALE_FACTOR = 1;
+export default async function* extractThumbnails(
```
mrjvs, line 7: first time I've ever seen a generator function used in the wild, exciting

frost768: Yeah, this was the first time I used it in a project. It really made sense to use it. Waiting for every thumbnail to load would be a bad ux

mrjvs (changes requested):
Nice job, just a few things missing.
I'm gonna retarget the PR to the v4 branch as its on the feature list for that version.

If you need assistance with the requested changes, we can also merge the incomplete PR and then I will finish it up.

src/utils/thumbnailCreator.ts
```diff
@@ -0,0 +1,51 @@
...
+  video.crossOrigin = "anonymous";
+
+  // Wait for the video metadata to load
+  await new Promise((resolve, reject) => {
+    video.addEventListener("loadedmetadata", resolve);
+    video.addEventListener("error", reject);
+  });
+
+  canvas.height = video.videoHeight * SCALE_FACTOR;
+  canvas.width = video.videoWidth * SCALE_FACTOR;
```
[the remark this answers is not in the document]
frost768: For this I will create an interface and implement it for hls and mp4 videos.
Hls one will have the same logic:
1. Create an instance of hls with the video source
2. Listen to corresponding metadata and seek events in hls js lib and seek to desired tims with hls controller
3. Convert fragment array data into base64 data url

Also an abortcontroller to cancel future requests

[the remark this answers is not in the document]
mrjvs: sounds good. A tip for the aborting. if you make it an internal video component. (which are just components of the video player that are always present) you can use useEffect to detect when the videoplayer has unloaded. Its also a good place to keep a video element always loaded (and hidden). a bit more efficient than making a new video element everytime.

mrjvs:
because the video player code will get a rewrite very soon for the new design, I'm gonna merge this now so the thumbnail component get roped in with the rewrite.

only thing missing was the design anyway, so good timing

mrjvs (approved):

PR #324, opened 2023-06-11
feat(provider): streamflix

lem6ns:
tested tv and movies, subtitles work on both
glorified 2embed

### DIFF
--- a/src/backend/providers/streamflix.ts
+++ b/src/backend/providers/streamflix.ts
@@ -1,0 +1,70 @@
+import { proxiedFetch } from "@/backend/helpers/fetch";
+import { registerProvider } from "@/backend/helpers/register";
+import {
+  MWCaptionType,
+  MWStreamQuality,
+  MWStreamType,
+} from "@/backend/helpers/streams";
+import { MWMediaType } from "@/backend/metadata/types";
+
+const streamflixBase = "https://us-west2-compute-proxied.streamflix.one";
+
+const qualityMap: Record<number, MWStreamQuality> = {
+  360: MWStreamQuality.Q360P,
+  540: MWStreamQuality.Q540P,
+  480: MWStreamQuality.Q480P,
+  720: MWStreamQuality.Q720P,
+  1080: MWStreamQuality.Q1080P,
+};
+
+registerProvider({
+  id: "streamflix",
+  displayName: "StreamFlix",
+  disabled: false,
+  rank: 69,
+  type: [MWMediaType.MOVIE, MWMediaType.SERIES],
+
+  async scrape({ media, episode, progress }) {
+    if (!this.type.includes(media.meta.type)) {
+      throw new Error("Unsupported type");
+    }
+
+    progress(30);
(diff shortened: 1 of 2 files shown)


src/backend/providers/streamflix.ts
```diff
@@ -0,0 +1,87 @@
...
+      const seasonNumber = media.meta.seasonData.number;
+      const episodeNumber = media.meta.seasonData.episodes.find(
+        (e) => e.id === episode
+      )?.number;
+
+      const streamRes = await proxiedFetch<any>(
+        `/api/player/tv?id=${media.tmdbId}&s=${seasonNumber}&e=${episodeNumber}`,
+        {
+          baseURL: streamflixBase,
+        }
```
thehairy, line 43:
```suggestion
      const streamRes = await proxiedFetch<any>(
        `/api/player/tv`,
        {
          baseURL: streamflixBase,
          params: {
              id: media.tmdbId,
              s: seasonNumber,
              e: episodeNumber
          },
        }
```

You can also just set the seasonNumber and episodeNumber to 1 if its a movie, or keep it undefined.
Then you dont need such a huge if/else with basically the same code repeating

thehairy (changes requested):

src/backend/providers/streamflix.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  async scrape({ media, episode, progress }) {
+    if (!this.type.includes(media.meta.type)) {
+      throw new Error("Unsupported type");
+    }
+
+    progress(30);
+    const type = media.meta.type === MWMediaType.MOVIE ? "movies" : "tv";
+    let seasonNumber: any;
+    let episodeNumber: any;
```
thehairy, line 35:
```suggestion
    let seasonNumber: string | undefined;
    let episodeNumber: string | undefined;
```

any just hurts my soul

src/backend/providers/streamflix.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const streamRes = await proxiedFetch<any>(`/api/player/${type}`, {
+      baseURL: streamflixBase,
+      params: {
+        id: media.tmdbId,
+        s: seasonNumber,
+        e: episodeNumber,
+      },
+    });
+    if (!streamRes.headers.Referer) throw new Error("No watchable item found");
```
thehairy, line 53: Are you sure thats a reliable way to tell if there is a video?
Haven't looked that deep into the provider.

thehairy (changes requested):

src/backend/providers/streamflix.ts
lem6ns: > Are you sure thats a reliable way to tell if there is a stream?
> Haven't looked that deep into the provider.

YES! If referer is blank, there is no streams/captions

src/backend/providers/streamflix.ts
lem6ns: Lol

thehairy (approved):

thehairy (approved):

mrjvs (approved):

PR #356, opened 2023-06-18
fix: language preference persistence

### LINKED ISSUE #320: [Bug]: Language doesnt get loaded after reload
### What happened?

my chosen language doesnt get loaded back in

Firefox

### Steps to reproduce?

1. choose german in settings
2. see that it changed to german
3. reload
4. its english again, but settings menu still says german

language gets stored in local storage, but does not get loaded

frost768:
resolves #320


mrjvs (approved):

mrjvs:
tested, works wonders.

Next time please tell us that youre picking up an issue (get assigned) before working on the code.
That makes sure we dont start working on the same issues at the same time.

PR #364, opened 2023-06-23
feat: human readable urls & quicksearch

### LINKED ISSUE #342: [Feature]: More readable URLs

More readable urls, for example: `/media/JW-1234--shrek-3`.

Allows people to see the title of the movie in the url, can be a good alternative instead of meta tags, which we aren't able to do due to being a client-only web application.

the title in the url would be purely cosmetic. it does not change any functionality or have any validation when loading it in.

### LINKED ISSUE #349: [Feature]: Quick search

A new endpoint: `/s/:query`.
it should search and immediately watch the first result.

Makes it quicker to search and watch.

_No response_

castdrian:
This pr closes #342 and also closes #349 by implementing the following:

- urls get the slugified media title appended, which is purely cosmetic
`/media/tmdb-show-103516` -> `/media/tmdb-show-103516-star-trek-strange-new-worlds`

- new endpoint /s/:query automatically searches and plays the first result for the provided query
`/s/smallville` -> `/media/tmdb-show-4604-smallville`


binaryoverload (approved):
This looks good to me, it just needs testing by us.

src/backend/metadata/getmeta.ts
```diff
@@ -224,10 +204,17 @@ export async function convertLegacyUrl(
...
+        strict: true,
+      })}`;
   }
 
   if (tmdbId) {
-    return `/media/tmdb-${type}-${tmdbId}`;
+    return `/media/tmdb-${type}-${tmdbId}-${slugify(meta.meta.title, {
+      lower: true,
+      strict: true,
+    })}`;
```
mrjvs, line 218: I think you should have a function for this instead. Something like `TMDBIdToUrlId(type, tmdbId, title)`.
You should then also use it in `TMDBMediaToId()`. So there is only one place that makes urls, makes it easier to change later

mrjvs (changes requested):
Just one tiny thing, for the rest it looks good. nice job.

src/backend/metadata/getmeta.ts
castdrian: good catch, will do :)

mrjvs (approved):

PR #367, opened 2023-06-23
v3.1.1

mrjvs:
changes:
 - fix poster not showing up in continue watching
 - updated documentation with TMDB
 - made TMDB configuration less error prone


mrjvs:
im gonna do a crime and merge without review :eyes:
whats the worst that can happen...

###### Famous last words

Issue #348, opened 2023-06-16
Feature freeze for version 4


mrjvs:
Hey all, we have entered a feature freeze for version 4.

No more PR's that aren't in the planning for v4 will be accepted. Not even translations or bug fixes.

You can check the planned features here: https://github.com/orgs/movie-web/projects/1
be aware, starting work on a feature/improvement/bugfix without communicating has a big chance of having your PR rejected.

In short:
 - only critical bugs that prevent normal use for large percentage of users will be merged into dev
 - only planned issues will be merged into v4
 - in case your PR doesn't fall under those conditions, it will be rejected.
 - when in doubt, ask us.

mrjvs:
No longer in a freeze

PR #375, opened 2023-06-29
Reduce requests to proxy by one

zisra:
This pull request replaces the [external_ids request](https://developer.themoviedb.org/reference/movie-external-ids) with an [append to response parameter](https://developer.themoviedb.org/docs/append-to-response)

I feel like this is pretty important because it reduced the amount of proxy bandwidth used + increases speed.


castdrian:
thonk, feature freeze + tmdb requests don't go through the proxy

zisra:
Oh, but I guess one request less is always better and keeps users waiting less (300ms)

castdrian:
true, still quite funny though

zisra:
If I had the time I would probably do a crazy calculation to determine how much time all users would save in a month

public/config.js
```diff
@@ -1,6 +1,6 @@
 window.__CONFIG__ = {
   // url must NOT end with a slash
-  VITE_CORS_PROXY_URL: "",
+  VITE_CORS_PROXY_URL: "https://rough.isra.workers.dev",
```
castdrian: leaked, should change the subdomain, and use the env file while deving

src/backend/metadata/tmdb.ts
```diff
@@ -155,10 +152,10 @@ export function getMediaDetails<
   TReturn = MediaDetailReturn<T>
 >(id: string, type: T): Promise<TReturn> {
   if (type === "movie") {
-    return get<TReturn>(`/movie/${id}`);
+    return get<TReturn>(`/movie/${id}?append_to_response=external_ids`);
   }
   if (type === "show") {
-    return get<TReturn>(`/tv/${id}`);
+    return get<TReturn>(`/tv/${id}?append_to_response=external_ids`);
```
castdrian, line 158:
```suggestion
    return get<TReturn>(`/tv/${id}`, { append_to_response: "external_ids" });
```

src/backend/metadata/types/tmdb.ts
```diff
@@ -169,6 +172,9 @@ export interface TMDBMovieData {
   video: boolean;
   vote_average: number;
   vote_count: number;
+  external_ids: {
+    imdb_id: string;
```
castdrian, line 176:
```suggestion
    imdb_id: string | null;
```

castdrian (changes requested):

castdrian (approved):

src/backend/metadata/types/tmdb.ts
zisra: Why? Causes linter errors too

castdrian: imdb id isn't guaranteed for shows, just fix the lint errors I guess

mrjvs:
Thanks for the PR, but it's not a planned issue, nor is it a critical bug fix.
Because of the feature freeze for v4, it wont be merged.

See #348 for more information, ive added some clarifying details.

PR #376, opened 2023-06-30
feat: multisearch & opensearch & sitelinks

### LINKED ISSUE #350: [Feature]: OpenSearch support + multi search

Add support for OpenSearch. It's the protocol that browsers use to make search engines addable to the browser bar.
MDN docs: https://developer.mozilla.org/en-US/docs/Web/OpenSearch

This will allow users to add the movie web search engine to their browser, for quick searching.

_No response_

castdrian:
this pr closes #350 by implementing:

- [x] multisearch
- [x] fix forgotten JW -> TMDB content type change
- [x] add ~dynamic~ opensearch support
- [x] add google sitelinks support

I tried to get the opensearch plugin to be dynamic for multiple hostnames but it seems to be impossible without SSR, unfortunately the browser doesn't care if it is modified client side


mrjvs:
Hmm, I dont really like hardcoding the domain. I'll research some compile template langs we can use.

That way we can disable the opensearch stuff for precompiled releases. And for compiled we can have an env var.

I'll get back to you on how to do it.

castdrian:
thank you I feel the same way, I tried for a couple hours with react and it was frustrating to say the least

src/backend/metadata/getmeta.ts
```diff
@@ -177,7 +182,16 @@ export async function convertLegacyUrl(
...
 
-  const mediaType = TMDBMediaToMediaType(type);
+  if (isLegacyMediaType(url)) {
+    const details = await getMediaDetails(id, TMDBContentTypes.TV);
+    return `/media/${TMDBIdToUrlId(
+      MWMediaType.SERIES,
+      details.id.toString(),
+      details.name
+    )}`;
+  }
```
mrjvs, line 192: Can you elaborate on why you switched from show to TV?

mrjvs (changes requested):
Didn't look at everything yet, just some initial thoughts

src/backend/metadata/getmeta.ts
castdrian: show was the JW content type, not the tmdb content type, was kinda overlooked during the refactor and ended up needing double conversion with multiseearch since it returns the content type, so I figured the correct action would be to actually make the tmdbcontent type type the actual values that tmdb uses instead of a double conversion

mrjvs: Fair, sounds logical

public/opensearch.xml
mrjvs, line 1: I've tested the opensearch and spaces become "+" characters. This is same behaviour as query params. so we should make a query param for search. You can just redirect it to /browse/:keyword immediately. and choose a seperate url for opensearch.

index.html
mrjvs, line 1: I've looked at templating engines, closest thing I could find is vituum: https://vituum.dev/
I think using handlebars is best. And instead of a pages directory you can put it in /src/assets. An asset folder is convention is preprocessed assets in these kinds of projects.

From experience, these kinds of builds steps can be tricky to configure. if you need help with it, let me know.

mrjvs (changes requested):
second batch, looked at everything this time :P

zisra:
Why are we using browse exactly? Search makes more sense to me because it does exactly what it says

mrjvs:
> Why are we using browse exactly? Search makes more sense to me because it does exactly what it says

because `/search/:type/:keyword` was the old one, so if you just change to `/search/:keyword`. all old backlinks wouldnt be the homepage anymore, it would search for `movie` or `series`.

having `/browse` would keep compatibility with old urls as we can redirect them properly

mrjvs:
We've discussed this with the maintainers and decided /browse is the best option

index.html
zisra: Wouldn't this be possible using react-helmet?

mrjvs: no, react-helmet does things on runtime. meta tags and structured data need to exist at compile time.

src/setup/App.tsx
```diff
@@ -82,11 +78,16 @@ function App() {
...
+                <Route exact path="/search/:type/:query?">
+                  <Redirect to="/browse/:query" />
+                </Route>
+                <Route exact path="/search/:type">
+                  <Redirect to="/browse" />
+                </Route>
+                <Route exact path="/browse/:query?" component={SearchView} />
+                <Route exact path="/">
+                  <Redirect to="/browse" />
+                </Route>
```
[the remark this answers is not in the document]
castdrian: could you elaborate a teensy tiny microscopic bit as to where it should point to, the original route config redirects root to search so I just copied that behaviour

index.html
castdrian: it seems like the vituum handlebars plugin is incompatible with the version of vite that the project uses, so I'm kinda stumped

src/setup/App.tsx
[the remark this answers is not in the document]
mrjvs: we're changing from how it originally was, too many people linked to /search/movie instead of main domain.

If no keyword is entered. it should be on `/`. if something is being searched, it should be on `/browse/:keyword`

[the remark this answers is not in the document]
mrjvs: You can do that with `<Route path={["/", "/browse/:keyword"]}>...</Route>`

castdrian:
oki doki, so it now reads APP_DOMAIN as env var and sets the opensearch accordingly, so should work with both dev and prod deployments, and localhost ofc

vite.config.ts
```diff
@@ -5,12 +5,14 @@ import { VitePWA } from "vite-plugin-pwa";
...
+import dotenv from 'dotenv';
+dotenv.config();
 
 export default defineConfig({
-  plugins: [
+  plugins:[
     handlebars({
       vars: {
-        message: "HELLOWORLD"
+        appdomain: process.env.APP_DOMAIN,
```
mrjvs: I recommend using vite's environment loading instead of dotenv.

```js
import { loadEnv } from "vite"

export default defineConfig(({ mode } => ({
  plugins:[
    handlebars({
      vars: {
        env: loadEnv(mode, process.cwd()),
      }
    })
  ],
  ...
}))
```

mrjvs: it would load all the env starting with VITE_. so VITE_X would make the object { x: ... }.
Just like we have with all the other environment variable config so far.

mrjvs (changes requested):

vite.config.ts
mrjvs: also please document it in an example.env

castdrian: oki doki, one and done

src/assets/opensearch.xml.hbs
```diff
@@ -0,0 +1,6 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>movie-web</ShortName>
+  <Description>The place for your favorite movies &amp; shows</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Url type="text/html" template="{{ appdomain }}/browse/{searchTerms}" />
```
mrjvs, line 5: The searchTerm needs to be in a query parameter. lets do `/browse/?q={searchTerms}`.
The q param should automatically redirect to `/browse/{queryHere}`.
You can add this redirection as a hook in the home view itself.

mrjvs: Same comment applies to the structured data for google search bar in the index.html

src/setup/App.tsx
```diff
@@ -82,12 +78,13 @@ function App() {
...
-                  path="/search/:type/:query?"
-                  component={SearchView}
-                />
-
+                <Route exact path="/search/:type/:query?">
+                  <Redirect to="/browse/:query" />
+                </Route>
+                <Route exact path="/search/:type">
+                  <Redirect to="/browse" />
+                </Route>
```
mrjvs, line 86: This redirection doesn't work. it just puts `:query` directly in the URL.

src/setup/App.tsx
```diff
@@ -82,12 +78,13 @@ function App() {
...
+                <Route path={["/", "/browse/:query"]} component={SearchView} />
```
mrjvs, line 87: I think I've said before to put the path as a list. but turns out that doesn't work.
Make it two seperate routes, rendering the same component. the one with `/` needs that have the `exact` prop.

vite.config.ts
mrjvs, line 1: This file now has a mix of tabs and spaces for indention. Please use only spaces, with size set to 2.

index.html
```diff
@@ -33,6 +33,26 @@
...
+		  "potentialAction": {
+			"@type": "SearchAction",
+			"target": {
+			  "@type": "EntryPoint",
+			  "urlTemplate": "https://movie-web.app/browse/{search_term_string}"
+			},
+			"query-input": "required name=search_term_string"
+		  }
+		}
+		</script>
```
mrjvs, line 55: This also needs to be toggled with the OPENSEARCH_ENABLED env.

index.html
```diff
@@ -33,6 +33,26 @@
     <meta name="referrer" content="no-referrer" />
 
     <title>movie-web</title>
+
+	<!-- OpenSearch -->
+    {{{ opensearch }}}
```
mrjvs, line 38: Please use handlebars if statement instead of injecting raw HTML.

See this documentation: https://handlebarsjs.com/guide/builtin-helpers.html#if

src/components/SearchBar.tsx
```diff
@@ -51,31 +34,6 @@ export function SearchBarInput(props: SearchBarProps) {
...
-            {
-              id: MWMediaType.SERIES,
-              name: t("searchBar.series"),
-              icon: Icons.CLAPPER_BOARD,
-            },
-          ]}
-          onClick={() => setDropdownOpen((old) => !old)}
-        >
-          {props.buttonText || t("searchBar.search")}
-        </DropdownButton>
```
mrjvs, line 77: the i18n key `searchBar.search` can now be removed from the locale files

src/setup/App.tsx
mrjvs, line 87: With the new url. search terms in the url seems to have broken.
previous behaviour:
 - on page load, it wil load in the query from the url and insert it into the searchbar.
 - on unfocus of the searchbar, the query that was typed in would be added to the url.

this functionality is still implemented, just needs to be adapted to work with the new url.

example.env
```diff
@@ -1,3 +1,5 @@
 # make sure the cors proxy url does NOT have a slash at the end
 VITE_CORS_PROXY_URL=...
 VITE_TMDB_READ_API_KEY=...
+VITE_OPENSEARCH_ENABLED=false
+VITE_APP_DOMAIN="http://localhost:5173"
```
mrjvs, line 5: please remove the quotes around the value and add a comment saying it should be without a slash at the end of the url. (similar to the cors proxy url comment

mrjvs (changes requested):
final batch of changes, I apologize for the amount of them.

src/components/SearchBar.tsx
mrjvs: nvm

mrjvs (approved):
Couldn't wait to work on my stuff today, so I decided to address the requested changes myself

castdrian:
haha alrighty-o excitement is fun

PR #404, opened 2023-08-15
Updated providers

mrjvs:

Not an issue, but it should be fixed anyway.
changes:
 - flixhq no longe ruses consumet API. just scrapes directly
 - upcloud decryption updated. fixes flixhq & gomovies
 - disabled broken providers
 - superstream now uses a faster cdn

> **Note**
> FlixHQ tv shows still do not work


binaryoverload (approved):
Unfortunately, I cannot find anything wrong with this PR

zisra:
Why have `type: [MWMediaType.MOVIE, MWMediaType.SERIES]` if only movies are supported?

Also, if I am not mistaken flixhq actually hit the tmdb endpoint of the consumet api which did not only run on flixhq but other sites. So we are removing providers

mrjvs:
> Why have `type: [MWMediaType.MOVIE, MWMediaType.SERIES]` if only movies are supported?
> 
> Also, if I am not mistaken flixhq actually hit the tmdb endpoint of the consumet api which did not only run on flixhq but other sites. So we are removing providers

tmdb endpoint of consumet API does not actually call any other providers. it just calls only flixHQ.
You can read their source code to verify.

the type it supports is kept the same as previously. as there are still plans to support it. I just couldnt get it work in time. FlixHQ does support series, just the code for it rn doesn't. Which is the same behaviour as previous version. series didnt work for flixHQ

Issue #335, opened 2023-06-16
[Bug]: Chromecasting unfinished


mrjvs:
### What happened?

- chromecasting doesnt receive subs
- chromecasting doesnt start at where the video left off, it starts at the beginning
- chromecasting doesnt support HLS

### What browsers are you seeing the problem on?

Firefox

FelixBoehme:
Will the casting code be affected by the planned video player refactoring?
If not I have a fix for the first two problems (haven't gotten around to the third issue).

I'd be happy to open a pull request if this can be used for the upcoming Version 4.

mrjvs:
This has been fixed in the latest version: #530

jrvgr:
Would like to reopen, Our chromecast unit isn't showing any. Would like to debug but no idea how

PR #419, opened 2023-08-31
Improved Chromecast integration

FelixBoehme:
This pull request resolves in parts #335.


src/video/components/actions/CaptionRendererAction.tsx
```diff
@@ -52,6 +53,7 @@ export function CaptionRendererAction({
   const videoTime = useProgress(descriptor).time;
   const { captionSettings, setCaptionDelay } = useSettings();
   const captions = useRef<ContentCaption[]>([]);
+  const casting = getPlayerState(descriptor).casting.isCasting;
```
castdrian, line 56:
```suggestion
  const { casting: { isCasting } } = getPlayerState(descriptor);
```

src/video/components/actions/ProgressAction.tsx
```diff
@@ -22,6 +22,7 @@ export function ProgressAction() {
   const controlRef = useRef<typeof controls>(controls);
   const [hoverPosition, setHoverPosition] = useState<number>(0);
   const [isThumbnailVisible, setIsThumbnailVisible] = useState<boolean>(false);
+  const casting = getPlayerState(descriptor).casting.isCasting;
```
castdrian, line 25:
```suggestion
  const { casting: { isCasting } } = getPlayerState(descriptor);
```

src/video/components/actions/ProgressAction.tsx
```diff
@@ -106,7 +107,7 @@ export function ProgressAction() {
           </div>
         </div>
       </div>
-      {isThumbnailVisible ? (
+      {isThumbnailVisible && !casting ? (
```
castdrian, line 110:
```suggestion
      {isThumbnailVisible && !isCasting ? (
```

src/video/components/actions/CaptionRendererAction.tsx
```diff
@@ -96,6 +98,7 @@ export function CaptionRendererAction({
     },
     []
   );
+  if (casting) return null;
```
castdrian, line 101:
```suggestion
  if (isCasting) return null;
```

castdrian (changes requested):

src/video/components/actions/ProgressAction.tsx
mrjvs: I dont agree

mrjvs: rename to isCasting is fine, no destructuring

src/video/state/providers/castingStateProvider.ts
```diff
@@ -140,18 +141,52 @@ export function createCastingStateProvider(
...
+      if (state.source?.caption?.url) {
+        const subtitles = new chrome.cast.media.Track(
+          1,
+          chrome.cast.media.TrackType.TEXT
+        );
+        subtitles.trackContentId = state.source?.caption?.url;
+        subtitles.trackContentType = "text/vtt";
+        subtitles.subtype = chrome.cast.media.TextTrackType.SUBTITLES;
+        subtitles.name = "Subtitles";
+        subtitles.language = "en";
```
mrjvs, line 160: not all of our captions we support are vtt. we use a library to parse various formats to VTT.
do you know if its possible to load a data url (`data:text/vtt;base64,...`) or something similar?

FelixBoehme: I'm not sure but I just realized when testing again, that this can't work. When testing I hosted it locally and used a fixed url for the subtitles to be able to test it, but i just realized, that it can't work in the way I imagined because the chromecast api has to have a url which can be loaded from the chromecast device which of course isn't possible with the blob approached used.
So sadly this won't work in the current state.
Don't know if I'll be able to come up with a fix for this, because simply using the the fetched urls won't always work if they're not in VTT format.
Sorry to disappoint, guess this feature will have to wait a bit longer.

mrjvs: the data url might work, you could give it a shot.

FelixBoehme: you mean the url before it is being parsed?

FelixBoehme:
I update it to use the url of the caption source, unfortunately the chromecast api seems to only support VTT and no other format.
So if the caption source is in the VTT format it does work, but otherwise not and it seems to be impossible to make it work unless the generated VTT file by you is hosted which I suspect you wouldn't want to do.
Hopefully if more subtitle providers are added this will hardly be an issue.

mrjvs:
even if the PR isn't fully featured, neither is the entire v4 branch.
so ill approve this if no major bugs come up in testing.

@binaryoverload, could you be a lad and test if casting still works as intended? (dont have a chromecast myself)

JamesHawkinss:
I've just tested this on my Chromecast-enabled device. Running into some issues:

* Subtitles only work on some movies, which is fine
* The Wolf of Wall Street seems to crash when scrubbing on the Chromecast device

No other major issues, I'm probably happy to merge

binaryoverload (approved):

binaryoverload:
@FelixBoehme please resolve the merge conflicts then I will be happy to merge this :)

FelixBoehme:
Im sorry I don't really know what to do this is my first PR and I have never resolved any conflicts how do I do this?
The button resolve conflicts is greyed out so I'm a little clueless on what to do now.

binaryoverload:
The best way is to merge the v4 branch into your branch locally, which will give you the merge issues to resolve. Then when those are resolved you can push again

castdrian:
merging branches isn't very clean though, ideally you always rebase instead

FelixBoehme:
good to know thanks, it should be resolved now I think

FelixBoehme:
Even though building the project failed I think I have fixed everything on my end but not sure. Hope this looks good now, let me know if there's anything I have missed and still need to fix

binaryoverload:
Not sure why it's not building, looks all good to me. Thanks again for the contribution @FelixBoehme

PR #578, opened 2023-12-23
Upgrade packages, bundling, performance

rafcontreras:
This pull request upgrades packages to latest versions, implements better bundling, improves performance


src/index.tsx
```diff
@@ -141,23 +141,23 @@ function TheRouter(props: { children: ReactNode }) {
...
-        </Suspense>
-      </HelmetProvider>
-    </ErrorBoundary>
-  </React.StrictMode>,
-  document.getElementById("root")
+const container = document.getElementById("root");
+const root = createRoot(container!);
+
+root.render(
+  <ErrorBoundary>
```
mrjvs, line 148: React.StrictMode was removed here. Is there a reason why? perhaps im missing some context.

src/pages/parts/auth/TrustBackendPart.tsx
```diff
@@ -54,16 +54,16 @@ export function TrustBackendPart(props: TrustBackendPartProps) {
         title={t("auth.trust.title")}
         icon={<Icon icon={Icons.CIRCLE_EXCLAMATION} />}
       >
-        <Trans i18nKey="auth.trust.host">
-          <span className="text-white">{{ hostname }}</span>
-        </Trans>
+        <span className="text-white">
+          <Trans i18nKey="auth.trust.host">{{ hostname }}</Trans>
+        </span>
```
mrjvs, line 59: Isn't this a breaking change for localization files?

The texts look like `Do you trust <0>{hostname}</0> for your backend?`. Doing it this way will make the `<0>` not work.

src/pages/parts/player/MetaPart.tsx
```diff
@@ -53,7 +53,7 @@ export function MetaPart(props: MetaPartProps) {
 
     let data: ReturnType<typeof decodeTMDBId> = null;
     try {
-      data = decodeTMDBId(params.media);
+      data = decodeTMDBId(params.media as string);
```
mrjvs, line 56: Why was this needed?

src/setup/App.tsx
```diff
@@ -65,63 +87,65 @@ function App() {
...
+        <Route
+          path="/media/:media/:season/:episode"
+          element={
+            <LegacyUrlView>
+              <Suspense fallback={null}>
+                <PlayerView />
+              </Suspense>
+            </LegacyUrlView>
+          }
+        />
```
mrjvs, line 116: Why were these pulled into two routes instead of one? They have identical contents.

mrjvs: Same question for /browse and /. (right below)

src/setup/App.tsx
```diff
@@ -65,63 +87,65 @@ function App() {
...
+        <Route path="/browse/:query?" element={<HomePage />} />
+        <Route path="/" element={<HomePage />} />
+        <Route path="/register" element={<RegisterPage />} />
+        <Route path="/login" element={<LoginPage />} />
+        <Route path="/about" element={<AboutPage />} />
 
         {shouldHaveDmcaPage() ? (
-          <Route exact path="/dmca" component={DmcaPage} />
+          <Route path="/dmca" element={<DmcaPage />} />
```
mrjvs, line 124: Are the exact props no longer needed? is it default?

vite.config.mts
```diff
@@ -14,6 +15,7 @@ export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd());
   return {
     plugins: [
+      million.vite({ auto: true }),
```
mrjvs, line 18: While building, I'm getting a lot of million source map warnings. Is there a way to fix that?

mrjvs (changes requested):
I agree with the premise of the PR but please communicate with us beforehand, we have an "approved" issue system for a reason. We want people to communicate these kinds of changes before acting on them.

src/index.tsx
rafcontreras: While on Dev it gives console errors on some libraries, will re-add it again.

src/pages/parts/auth/TrustBackendPart.tsx
rafcontreras: Will fix that.

src/pages/parts/player/MetaPart.tsx
rafcontreras: I was getting linting errors as params.media could be `string | null` and it is expecting string.

src/setup/App.tsx
rafcontreras: Routes no longer accept arrays of paths, so I had to break them up into separate routes.

src/setup/App.tsx
rafcontreras: That's right, in the documentation it says:

> You don't need to use an exact prop on <Route path="/"> anymore. This is because all paths match exactly by default. If you want to match more of the URL because you have child routes use a trailing * as in <Route path="users/*">.

vite.config.mts
rafcontreras: They are only warnings, while opening source maps in the browser all seems ok.

mrjvs: It does seem like the output is fine, I've mentioned this case in an existing bug report: https://github.com/aidenybai/million/issues/844

We can keep it like this for now, when it's hopefully fixed in the next version of million, we can just update.

src/setup/App.tsx
mrjvs: Oh cool, good change on their end.

src/pages/parts/player/MetaPart.tsx
mrjvs: It seems they changed the type of useParams then.
Instead of casting, I'd rather have a explicit check for it that just throws if media param isn't set. (inside the useAsync)

src/setup/App.tsx
mrjvs: Hmm, annoying change. but code should be good then. We can keep it like this.

src/index.tsx
rafcontreras: Re-added StrictMode

src/pages/parts/auth/TrustBackendPart.tsx
rafcontreras: I've modified the code.

src/pages/parts/player/MetaPart.tsx
rafcontreras: Now it throws an error if the param isn't set

src/pages/parts/auth/TrustBackendPart.tsx
mrjvs: Seems like it's still not resolved.
The only white part should be the hostname, perhaps you can just revert it to what it was orginally?

mrjvs: Decided to fix it myself, cuz I wanted to play around with some of the bundling myself anyway.
Fixed in latest commit

mrjvs (approved):

mrjvs:
Thanks for your PR, im sure our users will appreciate the optimizations :)

PR #597, opened 2023-12-27
CDN replacements

mrjvs:
changes:
 - allow option of overwriten cdn domains with a custom one specified in config.

### DIFF
--- a/src/setup/config.ts
+++ b/src/setup/config.ts
@@ -95,6 +95,6 @@ export function conf(): RuntimeConfig {
           .map((s) => s.trim())
           .filter((s) => s.length > 0),
       )
-      .filter((v) => v.length > 0), // The format is <beforeA>:<afterA>,<beforeB>:<afterB>
+      .filter((v) => v.length === 2), // The format is <beforeA>:<afterA>,<beforeB>:<afterB>
   };
 }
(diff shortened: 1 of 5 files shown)


src/setup/config.ts
```diff
@@ -84,5 +87,14 @@ export function conf(): RuntimeConfig {
...
       .filter((v) => v.length > 0), // Should be comma-seperated and contain the media type and ID, formatted like so: movie-753342,movie-753342,movie-753342
+    CDN_REPLACEMENTS: getKey("CDN_REPLACEMENTS", "")
+      .split(",")
+      .map((v) =>
+        v
+          .split(":")
+          .map((s) => s.trim())
+          .filter((s) => s.length > 0),
+      )
+      .filter((v) => v.length > 0), // The format is <beforeA>:<afterA>,<beforeB>:<afterB>
```
binaryoverload, line 98: This should be checking if there are two elements in each array perhaps?

```suggestion
      .filter((v) => v.length === 2), // The format is <beforeA>:<afterA>,<beforeB>:<afterB>
```

binaryoverload (changes requested):

binaryoverload (approved):

PR #604, opened 2023-12-28
Add playlist download filename

robertsspaceindustries:
Playlists were apparently playing in-browser for some users, adding the [download attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) should fix it. Issue discussed on the Discord end.

### DIFF
--- a/src/components/player/atoms/settings/Downloads.tsx
+++ b/src/components/player/atoms/settings/Downloads.tsx
@@ -43,6 +43,7 @@ export function DownloadView({ id }: { id: string }) {
   const { t } = useTranslation();
   const downloadUrl = useDownloadLink();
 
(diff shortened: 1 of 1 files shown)


JipFr:
Have you tested this at all? IIRC the download attribute is only allowed on same-domain links. 

We usually just link to the source so people can download it whichever way they want, which is usually a "download" button provided by the browser.

robertsspaceindustries:
> Have you tested this at all? IIRC the download attribute is only allowed on same-domain links.
> 
> We usually just link to the source so people can download it whichever way they want, which is usually a "download" button provided by the browser.

Yes and no, the issue in question doesn't happen for me.

Issue #573, opened 2023-12-22
[Feature]: Skip forward buttons


sooswastaken:
### What feature do you want to add?

Customization of the skip forward buttons (change from 1s to custom amount like 10 or 5s)

### Why do you want to have this feature?

More customizability

DenuxPlays:
Maybe add  "youtube-style" like shortcuts.
Arrow-keys => 5sec 
And J and L => 10sec

JipFr:
What button has 1s now?

blikje7up:
arrow key already is 5 seconds

qtchaos:
I have implemented changes that add the YouYube-style shortcuts (`j`, `l` (skip by 10 sec) and `,`, `.` (skip by 1 sec). Will PR once i get assigned :p

PR #612, opened 2023-12-28
Improve video progress/playback keybinds

### LINKED ISSUE #577: [Feature]: Seek using numbers [0-9]

Seeking through the video player using the number keys

Many other sites support it

_No response_

qtchaos:
This pull request resolves ~~#577~~ and #573.

Implemented *most* of the [YouTube player shortcuts](https://support.google.com/youtube/answer/7631406), the only outliers being the `,` and `.` keys which skip backward and forward by a frame respectively, since I don't believe we can skip the video with that much granularity I decided these should instead use 1 second.

Adds the following shortcuts in the player:
 * ~~`0` Seek to the beginning of the media.~~
 * ~~`1-9` Seek to the 10% to 90% of the media.~~
 * `J` Seek backward 10 seconds in player.
 * `l` Seek forward 10 seconds in player.
 * `,` While the video is paused, skip forward 1 second.
 * `.` While the video is paused, skip backward 1 second.
 * `<` Decrease playback speed
 * `>` Increase playback speed


binaryoverload:
Hi,

Thanks for doing this. As mentioned on the now closed #577, could you please remove the 0-9 skipping shortcuts as they make it too easy to lose your progress

binaryoverload:
Hi,

Thanks for doing this. As mentioned on the now closed #577, could you please remove the 0-9 skipping shortcuts as they make it too easy to lose your progress

qtchaos:
Removed seeking numbers in https://github.com/movie-web/movie-web/pull/612/commits/27b28a6d96b4742182f4cb96841e6b52c1110712 👍

src/components/player/internals/KeyboardEvents.tsx
```diff
@@ -83,11 +91,28 @@ export function KeyboardEvents() {
         );
       if (k === "m") dataRef.current.toggleMute();
 
+      // Video playback speed
+      if (k === ">" || k === "<") {
+        const options = [0.25, 0.5, 1, 1.5, 2];
+        const idx = options.indexOf(dataRef.current.mediaPlaying?.playbackRate);
+        const nextIdx = idx + (k === ">" ? 1 : -1);
```
binaryoverload, line 98: If the existing index doesn't exist (Which may be possible if we change the settings implementation in the future) then we should handle that

I suggest defaulting the index to the index of `1` (2 in this case) if the index is -1

src/components/player/internals/KeyboardEvents.tsx
```diff
@@ -83,11 +91,28 @@ export function KeyboardEvents() {
...
       // Video progress
       if (k === "ArrowRight")
         dataRef.current.display?.setTime(dataRef.current.time + 5);
       if (k === "ArrowLeft")
         dataRef.current.display?.setTime(dataRef.current.time - 5);
+      if (k === "j")
+        dataRef.current.display?.setTime(dataRef.current.time - 10);
+      if (k === "l")
+        dataRef.current.display?.setTime(dataRef.current.time + 10);
+      if (k === "." && dataRef.current.mediaPlaying?.isPaused)
```
binaryoverload, line 112: It might be a good idea to pause when you press the keybind (I believe this is the behaviour of YouTube). Thoughts?

binaryoverload (changes requested):

src/components/player/internals/KeyboardEvents.tsx
qtchaos: YouTube doesn't pause the video when using the `j` or `l` shortcuts, If we would pause the video when using them. it would probably reduce playback lag when pressed multiple times in a row. If you think that might be better then I'll implement those changes, but I really have no direct opinion.

binaryoverload: Oh I see it only works on YT when paused - that's fine then

src/components/player/internals/KeyboardEvents.tsx
qtchaos: Added these changes in https://github.com/movie-web/movie-web/pull/612/commits/5e52ea8e8fdcea482be57b75ee94c46c00c5e33d

zisra:
Keybinds should somehow be documented, maybe through the discord bot, FAQ page, or docs

PR #619, opened 2023-12-29
Migrate libraries: ISO-639-1 -> locale-code

GustavoMelloGit:
This pull request resolves the issue where the ISO-639-1 library has a very limited number of languages, excluding variants such as pt-BR.


src/assets/locales/pt-BR.json
mrjvs, line 1: Please do not modify the translations manually, use the community translation tool. Check out contributing guidelines for more info.

src/assets/languages.ts
```diff
@@ -21,28 +22,30 @@ import vi from "@/assets/locales/vi.json";
...
+  "ar-SA": ar,
+  "es-ES": es,
+  "et-EE": et,
   tok,
 };
+
 export type Locales = keyof typeof locales;
 
-export const rtlLocales: Locales[] = ["he", "ar"];
+export const rtlLocales: Locales[] = ["he-IL", "ar-SA"];
```
mrjvs, line 51: This is breaking change. People that already have a selected language selected will be changed back to english. Plus the preferred language stored on the account would be wrong.

If we're going to support alpha-2 iso codes. It will need to be backwards compatible.

Not to mention that subtitles from provider package will still be using the other codes, so subtitles will likely break as well then.

mrjvs (changes requested):

src/assets/languages.ts
GustavoMelloGit: > Not to mention that subtitles from provider package will still be using the other codes, so subtitles will likely break as well then.

What do you propose to fix that? I have no idea what to do...

binaryoverload:
Due to the complexities of implementing this feature, we have decided that the core team will do it ourselves. Apologies for wasting your time - thank you for your effort!

PR #673, opened 2023-12-30
fix: movies with unavailable posters will be shown in the last row

### LINKED ISSUE #665: [Bug]: Movies with missing posters displayed in search results
### What happened?

Searching for movies sometimes leads to entries with missing posters. Clicking on these movies often generates a “result not found” message. This creates a bad user experience.

Firefox

### Steps to reproduce?

1. Go to the search page
2. Search nemo (You found there were two movies that have no poster)
3. click anyone (When you click on those, the result is 'not found')

image
image

ashudevcodes:
This pull request resolves #665 

# All movies with unavailable posters will be shown in the last row
image
image


src/backend/metadata/search.ts
```diff
@@ -24,6 +24,17 @@ export async function searchForMedia(query: MWQuery): Promise<MediaItem[]> {
...
+  results.sort((a, b) => {
+    if (a.poster === undefined) {
+      return 1;
+    }
+    if (b.poster === undefined) {
+      return -1;
+    }
+    return 0;
+  });
+
```
qtchaos, line 36: pnpm lint:fix

src/backend/metadata/search.ts
```diff
@@ -24,6 +24,17 @@ export async function searchForMedia(query: MWQuery): Promise<MediaItem[]> {
...
 
-  cache.set(query, results, 3600); // cache results for 1 hour
+  results.sort((a, b) => {
+    if (a.poster === undefined) {
+      return 1;
+    }
+    if (b.poster === undefined) {
+      return -1;
+    }
+    return 0;
```
mrjvs, line 34: This destroys the order that the undefined posts come in.
Especially since the sequence of when the elements get compared is undefined behaviour, so sorting would come out randomized in theory.

This fix probably can't be done with a sort(), so I recommend making two seperate arrays (with poster and without poster) using filter(). then concatinating them.

mrjvs (changes requested):

src/backend/metadata/search.ts
binaryoverload: Sorting shouldn't come out as randomised since Array.sort is a stable sort surely?

binaryoverload: After a discussion internally, please proceed with JVS's suggestion of filtering @ashudevcodes

src/backend/metadata/search.ts
```diff
@@ -24,6 +24,12 @@ export async function searchForMedia(query: MWQuery): Promise<MediaItem[]> {
     return formatTMDBMetaToMediaItem(formattedResult);
   });
 
-  cache.set(query, results, 3600); // cache results for 1 hour
-  return results;
+  const movieWithposters = results.filter((movie) => movie.poster);
+  const movieWithoutposters = results.filter((movie) => !movie.poster);
+
+  const sortedresult = movieWithposters.concat(movieWithoutposters);
```
binaryoverload, line 30: Minor stylistic nit-pick

```suggestion
  const movieWithPosters = results.filter((movie) => movie.poster);
  const movieWithoutPosters = results.filter((movie) => !movie.poster);

  const sortedresult = movieWithPosters.concat(movieWithoutPosters);
```

binaryoverload (changes requested):

mrjvs (approved):

Issue #672, opened 2023-12-30
[Feature]: Move to the top of the page with every new search.


zaim-abbasi:
### What feature do you want to add?

Whenever we navigate to the end of the movies grid from the Search page and search for another movie while at the bottom of the page, our search results are shown at the end of the page instead of bringing us to the top.

If you can't understand my issue, you can:
* Search for any movie/show.
* Scroll to the middle or end of the page.
* Now, search for another movie while you are in the middle of the page.
* The new search results will be displayed, and you will find yourself in the middle or at the end of the page where you left off on the last search. Instead, it should bring us to the top of the page with every new search.

### Why do you want to have this feature?

Increases Usability

### Anything other details to share?

It also seems to be a bug ig

mrjvs:
cannot reproduce, will always jump to top

PR #674, opened 2023-12-30
Solved  [Feature]: Move to the top of the page with every new search. #672

Ayushmaanagarwal1211:
This pull request resolves #XXX

### DIFF
--- a/example.env
+++ b/example.env
@@ -1,8 +1,0 @@
-VITE_TMDB_READ_API_KEY=...
-VITE_OPENSEARCH_ENABLED=false
-
-# make sure the cors proxy url does NOT have a slash at the end
-VITE_CORS_PROXY_URL=...
-
-# make sure the domain does NOT have a slash at the end
-VITE_APP_DOMAIN=http://localhost:5173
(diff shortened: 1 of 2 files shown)


JipFr:
I don't really agree with this. I feel like even if the issue was approved, which it hasn't been, it shouldn't scroll past the title so you can see the top results.

Ayushmaanagarwal1211:
> I don't really agree with this. I feel like even if the issue was approved, which it hasn't been, it shouldn't scroll past the title so you can see the top results.

so if i add a feature is that when user unfocused from search bar then it moves to the top otherwise not .

JipFr:
IMO when something is searched it should scroll to the title, but I want to hear Jvs' thoughts on this before anything else is done. This discussion should take place in the issue, not this PR (imo)

tan-ahmed:
this is a good UX thing

zaim-abbasi:
Or maybe we could  scroll to a position some pixels above the first search result image, subtracting a specific number of pixels from the offsetTop value of the first search result. like this:

const firstResult = document.getElementById('firstSearchResult');

  if (firstResult) {

    const offsetAboveResult = 20; // 20 pixels above the first search result

    // Calculate the offsetTop
    const offsetTop = firstResult.offsetTop;

    window.scrollTo({
      top: offsetTop - offsetAboveResult,
      left: 0,
      behavior: 'smooth',
    });
  }

mrjvs:
As a heads up, I won't have the time to review properly until after new years.

mrjvs:
movie-web currently already goes to top of the page when searching, there is no need for this PR.
additionally, the issue was not approved to begin with.

PR #684, opened 2023-12-31
change fonts to bunny

coldrdy:
change fonts to bunny


mrjvs:
Pr doesn't have template, nor is this an approved issue that you were assigned to. Closing

PS, we don't think changing font cdns is necessary

PR #721, opened 2024-01-05
Implement MediaSession support

### LINKED ISSUE #176: [Feature]: MediaSession implementation

qtchaos:
This pull request resolves #176

Adds all video related MediaSession functionality.

Tried to limit updating media progress as much as possible, not sure if this is necessary or not, but thought it would be nice. This unfortunately complicates the code a bit, so if this isn't necessary I'll remove it.

~~Not that happy with the structure of the MediaSession component, i guess I'll see what reviews think :)~~


src/components/player/internals/MediaSession.ts
```diff
@@ -0,0 +1,178 @@
...
+      updatePositionState(e.seekTime);
+    });
+
+    if (dataRef.current.meta?.episode?.number !== 1) {
+      navigator.mediaSession.setActionHandler("previoustrack", () => {
+        changeEpisode(-1);
+      });
+    } else {
+      navigator.mediaSession.setActionHandler("previoustrack", null);
+    }
```
mrjvs, line 158: for movies, this will also show the previoustrack button. since episode is null and `null !== 1` evaluates true

src/components/player/internals/MediaSession.ts
```diff
@@ -0,0 +1,178 @@
...
+        dataRef.current.progress.time + skipTime,
+      );
+      updatePositionState(dataRef.current.progress.time + skipTime);
+    });
+
+    navigator.mediaSession.setActionHandler("seekto", (e) => {
+      if (!e.seekTime) return;
+      dataRef.current.display?.setTime(e.seekTime);
+      updatePositionState(e.seekTime);
+    });
```
mrjvs, line 150: seems like if you both the seek forward/backward and seekto. the progress bar doesn't show. So maybe only implement seekto https://stackoverflow.com/a/76090643

mrjvs (changes requested):
On top of the few bugs I found. There also seems to be a lack of event listening on the clients end. For example, how does the mediasession state get updated if the user pauses on our player? Using useEffect, you should be able to listen to changes to our state and update the mediasession accordingly.

mrjvs:
Play pause seems to work nicely, but next/prev buttons seem to not update the state properly.
It goes to the next episode, but your components thinks its still the old one. So next episode again won't trigger it correctly. Also the url doesn't update when switching episode. See gif for reproduction:

[!https://i.imgur.com/6y4N2dW.gif](https://i.imgur.com/6y4N2dW.gif)

binaryoverload (approved):

mrjvs:
Holy shit, it got merged :tada:

PR #722, opened 2024-01-05
Push Docker image to registry after release

### LINKED ISSUE #675: [Feature]: Docker image for frontend

It would be great if there was also a simple docker image to serve the frontend with nginx for example.

Deploying and updating the movie-web frontend would take me 2 seconds if I just had a Docker image with the latest version of the movie-web frontend. But with the current situation I have to download the zip file, change manually the configuration file instead of just using env variables and whenever I want to update the frontend I need to go download the latest release instead of having a Docker image of it. Moreover, seeing as I don't want to use an external service to host my website but my own Traefik instance to serve the website, I will need to make myself a nginx container to serve the static files instead of having a community configured one.

If you agreed to have and maintain that, I could make a PR to add the Dockerfile and the pipeline to deploy the built docker image to the registry.

sweenu:
This pull request resolves #675

 - [ ] I have read and agreed to the [code of conduct](https://github.com/movie-web/movie-web/blob/dev/.github/CODE_OF_CONDUCT.md).
 - [ ] I have read and complied with the [contributing guidelines](https://github.com/movie-web/movie-web/blob/dev/.github/CONTRIBUTING.md).

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -13,6 +13,6 @@ RUN pnpm run build
 
 # production environment
 FROM nginx:stable-alpine
-COPY --from=build /app/build /usr/share/nginx/html
+COPY --from=build /app/dist /usr/share/nginx/html
 EXPOSE 80
 CMD ["nginx", "-g", "daemon off;"]
--- a/.github/workflows/deploying.yml
+++ b/.github/workflows/deploying.yml
@@ -127,3 +127,52 @@ jobs:
         asset_path: ./movie-web.zip
         asset_name: movie-web.zip
         asset_content_type: application/zip
+
+  registry:
+    name: Push to registry
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    env:
+      REGISTRY: ghcr.io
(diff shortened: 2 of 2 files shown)


mrjvs:
@weeryan17

sweenu:
I'd like to change the CMD to an ENTRYPOINT to a script that will change the config file from the environment variables and then run nginx, but I wanted to know first if there was a way for the application to be able to read from environment variables directly the way it does it when we deploy with Vercel. That would be way cleaner.

dockerfile
```diff
@@ -6,13 +6,13 @@ RUN corepack enable
 
 COPY package.json ./
 COPY pnpm-lock.yaml ./
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install
```
binaryoverload, line 9: I think we should keep this, as this is a CI specific flag that won't be enabled by default as we're not running it in the GH Action directly

binaryoverload (changes requested):

binaryoverload (approved):

PR #723, opened 2024-01-06
fix Brazil flag

### LINKED ISSUE #680: [Bug]: Wrong flag in pt-br translation
### What happened?

Recently, Brazilian Portuguese translation was added, but the flag that appears in the options menu is that of European Portuguese, which is a significantly different language.

Chrome

### Steps to reproduce?

Go to the language selector and check the flag

I have already implemented the solution; I just need permission to submit the pull request.

QSRafael:
This pull request resolves #680

### DIFF
--- a/src/utils/language.ts
+++ b/src/utils/language.ts
@@ -12,7 +12,7 @@ const countryPriority: Record<string, string> = {
   nl: "nl",
   fr: "fr",
   de: "de",
-  pt: "pt",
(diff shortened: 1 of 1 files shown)


binaryoverload:
Thanks for the PR, however this doesn't fit the issue in the way we'd like. I will be making a PR shortly that changes the system we have to get the region from the language code itself.

PR #726, opened 2024-01-06
Language code fixes

### LINKED ISSUE #680: [Bug]: Wrong flag in pt-br translation

binaryoverload:
This pull request resolves #680

This changes our language utils to read the region from the language code - this means we only need to specify manual overrides for languages that either:
 a) have an incorrect region
 b) don't have a region in the language code (i.e. `zh-Hant`)


mrjvs:
Ew, force push

src/utils/language.ts
```diff
@@ -5,19 +5,11 @@ const languageOrder = ["en", "hi", "fr", "de", "nl", "pt"];
...
-  en: "us",
-  nl: "nl",
-  fr: "fr",
-  de: "de",
-  pt: "pt",
-  ar: "sa",
-  es: "es",
   zh: "cn",
-  ko: "kr",
-  ta: "lk",
```
mrjvs, line 20: Does this work? One of the issues that this mapping solved was that for example in NL. The Netherlands isn't the first country on the list of countries that speak the language, the first one would be Belgium since its earlier in alphabetical order.

The mapping wasn't just for stuff like scripts

mrjvs (changes requested):

src/utils/language.ts
binaryoverload: Yes it does because we have the default language codes which have Dutch as `nl-NL` - the `NL` being the region

binaryoverload: If we didn't have the default language codes you would be right, but because we have a mapping of language -> language with region then it works

mrjvs: Ah right, can you double check if subtitle flags still work then? I'm unsure if the default region mapping is applied there

binaryoverload: Fair point well made - they didn't (Because I forgot this logic was used there. I've updated the getCountryCodeForLocale method to use the populateLanguageCode method

mrjvs (approved):

PR #807, opened 2024-01-23
feat: add ide extension for better i18n handling in vs code

NilsJacobsen:
> Let me know what I need to change in order to get this merged

## Improve i18n handling in vs-code with an [ide extension](https://inlang.com/m/r7kp499g)

**Improves i18n tasks:**
- Inline Annotations
- Lint messages
- Extract messages
- Update Translations

## What I added

- project.inlang/settings.json as configuartion
- I added it to vs-code recommendations so other devs can use it too
- Project id file was generated by the [inlang cli](https://inlang.com/m/2qj2w8pu) as a validated the project setup (it's for monorepo support)

- [ ] I have tested all of my changes. (I validated the inlang project, shouldn't affect tests)


NilsJacobsen:
@castdrian you are right. Updated it.

I still think the extension could be beneficial for the project. Feel free to close if you disagree.

binaryoverload:
Does the extension provide autocompletion for language strings within our React files?

binaryoverload:
It's not a problem per se, but it would be handy.

If the extension doesn't do that I don't think it will benefit us, as we don't manually edit languages files that often and only the English one at that

PR #819, opened 2024-01-24
Small extension fixes

mrjvs:
changes:
 - specific text per browser for extension setup screen
   - chrome: show chrome extension install link
   - firefox: show firefox extension install link + extra firefox help
   - ios: disclaimer text that ios doesnt have extensions
   - unknown: show both chrome and firefox install links + extra firefox help
 - Fix bug where first load the page couldn't talk to extension
 - Fix too short of a timeout when checking for proxy response
 - Move start of bolding for HLS disclaimer
 - Fix stackoverflow error when opening download screen on lower ram browsers

### DIFF
--- a/src/utils/detectFeatures.ts
+++ b/src/utils/detectFeatures.ts
@@ -1,3 +1,4 @@
+import { detect } from "detect-browser";
 import fscreen from "fscreen";
 import Hls from "hls.js";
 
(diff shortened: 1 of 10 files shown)


src/utils/detectFeatures.ts
```diff
@@ -52,3 +53,26 @@ export function canPlayHlsNatively(video: HTMLVideoElement): boolean {
...
+
+  // not a browser or failed to detect
+  if (res?.type !== "browser") return "unknown";
+
+  if (res.name === "ios" || res.name === "ios-webview") return "ios";
+  if (
+    res.name === "chrome" ||
+    res.name === "chromium-webview" ||
+    res.name === "edge-chromium"
+  )
```
binaryoverload, line 74: Maybe add `opera` into this mix?

binaryoverload (changes requested):

src/utils/detectFeatures.ts
mrjvs: I thought opera wasn't one of the options. But I'll check

binaryoverload: It is, see here: https://github.com/DamonOehlman/detect-browser/blob/master/src/index.ts#L85

binaryoverload (approved):

PR #825, opened 2024-01-25
Created an 'x' button to clear search query

### LINKED ISSUE #824: [Feature]: Create an 'x' button to clear search query

A 'x' button to the right side of the search bar to clear the query.

Goes without saying, but this is an essential QoL feature.

I am in the process of creating a PR to address this issue.

RyloRiz:
This pull request resolves #824

 - [x] I have read and agreed to the [code of 
conduct](https://github.com/movie-web/movie-web/blob/dev/.github/CODE_OF_CONDUCT.md).
 - [x] I have read and complied with the [contributing 
guidelines](https://github.com/movie-web/movie-web/blob/dev/.github/CONTRIBUTING.md).


src/components/form/SearchBar.tsx
```diff
@@ -57,6 +57,15 @@ export const SearchBarInput = forwardRef<HTMLInputElement, SearchBarProps>(
             className="w-full flex-1 bg-transparent px-4 py-4 pl-12 text-search-text placeholder-search-placeholder focus:outline-none sm:py-4 sm:pr-2"
             placeholder={props.placeholder}
           />
+
+          <div
+            onClick={() => {
+              setSearch("");
+            }}
+            className="cursor-pointer absolute bottom-0 right-5 top-0 flex max-h-14 items-center text-search-icon"
```
mrjvs, line 65: Few things regarding the look and feel of the button:
1. It needs a hover effect and an mousedown effect (active). lets do:
    - Show a circular background on hover, make the icon white
    - When holding down mouse click, add a `scale-110`.
2. Hitbox is currently super thin but stretched from top to bottom. make the hitbox this instead:

src/components/form/SearchBar.tsx
```diff
@@ -57,6 +57,15 @@ export const SearchBarInput = forwardRef<HTMLInputElement, SearchBarProps>(
             className="w-full flex-1 bg-transparent px-4 py-4 pl-12 text-search-text placeholder-search-placeholder focus:outline-none sm:py-4 sm:pr-2"
             placeholder={props.placeholder}
           />
+
+          <div
+            onClick={() => {
+              setSearch("");
```
mrjvs, line 63: Because this unfocuses first, then resets the search input (on desktop with mouse), it will not update the URL after clearing. Fix it by triggering the onfocus event manually after clearing.

src/components/form/SearchBar.tsx
```diff
@@ -57,6 +57,15 @@ export const SearchBarInput = forwardRef<HTMLInputElement, SearchBarProps>(
             className="w-full flex-1 bg-transparent px-4 py-4 pl-12 text-search-text placeholder-search-placeholder focus:outline-none sm:py-4 sm:pr-2"
             placeholder={props.placeholder}
           />
+
+          <div
```
mrjvs, line 61: Let's make this only show if something is actually typed into the field, it's weird to have an clear button when you haven't done anything yet.

mrjvs (changes requested):

RyloRiz:
@mrjvs All requests have been implemented & tested on both desktop and mobile, please let me know if there are any issues.

themes/default.ts
```diff
@@ -160,6 +160,7 @@ export const defaultTheme = {
       // search bar
       search: {
         background: tokens.shade.c500,
+        hoverBackground: tokens.shade.c600,
```
mrjvs, line 163: Make sure to add this to all color themes, not just default. It will be just a copy paste into those other files, so not a lot of effort.

mrjvs (changes requested):
I appreciate the effort, but this is kinda overcomplicated.
Here are ways to make it simpler, that'd id like to see in this PR:

For the hiding of the clearbutton, use conditional rendering: https://react.dev/learn/conditional-rendering
The condition can just be `props.value.length > 0`.

For the url fixes, all you needed to do is set this as onclick handler:
```ts
onClick={() => {
  setSearch("");
  props.onFocus();
}}
```

For the hover effects, you can just put the text colors on the clickable hitbox of the div. no need to use groups to dig it down further. The icon inherits the text color from it's parents.

With those approaches, you wont need useEffect, useRef or need to add another state variable.

RyloRiz:
> For the url fixes, all you needed to do is set this as onclick handler:

As `setSearch/setState` is an asynchronous process, the `search` variable does not update in time for `props.onUnFocus` to properly update the URL. The current system using these two lines as the clear button's `onClick` works as follows:

```ts
setSearch("");
props.onUnFocus();
```

*Assume the current search term is **"test"***:

* `TextInputControl`'s `onBlur` handler fires `onUnFocus`, which sets the URL to the current (old) query (the textinputcontrol was unfocused in favor of the clear button)
* `setSearch` fires with **no** URL-updating permission (`commitToUrl = false`) and the intent to update the query to `""`
* `props.onUnFocus` fires immediately after **with** URL-updating permission (`commitToUrl = true`), *but* `search` has not updated yet due to the asynchronous nature of `setSearch/setState` and instead this re-applies the old query as the new query

A possible fix is to edit the `onUnFocus` function declaration in the `useSearchQuery` hook to allow for an optional parameter that would use it instead of the current value of `search` and pass that along to the `updateParams` call that *does* have the permission to update the URL. I can confirm through testing that this method does work as intended.

All other concerns outlined in your reply have been fixed without issues.

mrjvs:
> A possible fix is to edit the `onUnFocus` function declaration in the `useSearchQuery` hook to allow for an optional parameter that would use it instead of the current value of `search` and pass that along to the `updateParams` call that _does_ have the permission to update the URL. I can confirm through testing that this method does work as intended.

I like this method, let's go with that!

mrjvs (approved):

PR #827, opened 2024-01-25
Add FormData, URLSearchParams and User-Agent support for extension

### LINKED ISSUE #820: [Bug]: Extension fetching limitations
### What happened?

There are a few limitations that prevent extension from working fully:
- Instance objects like FormData and UrlSearchParams can't be sent over messaging framework. It will fail to make the request with a null response.
- User-Agent header cannot be set with the request

These two are making showbox/febbox currently broken for extension scraping.

Chrome

### Steps to reproduce?

_No response_

FormData and UrlSearchParams can be represented in plain objects, just to add a bodyType. parameter or smth to parse it back after transfering to extension.

For the user agent, maybe XMLHttpRequest will work. Or maybe using declarativeNetRequest we can set the user agent before we make our own API request.

JorrinKievit:
This pull request resolves #820 

image


src/backend/extension/request.ts
```diff
@@ -0,0 +1,22 @@
+import { ExtensionMakeRequestBodyType } from "./plasmo";
+
+export function getBodyTypeFromBody(
+  body: unknown,
+): ExtensionMakeRequestBodyType {
+  if (typeof body === "string") return "string";
+  if (body instanceof FormData) return "FormData";
+  if (body instanceof URLSearchParams) return "URLSearchParams";
+  if (typeof body === "object") return "object";
+  return "string";
```
mrjvs, line 10: I think we can safely default to object instead of string. then the last check isnt needed either

src/backend/extension/request.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export function convertBodyToObject(body: unknown): any {
+  if (body instanceof FormData || body instanceof URLSearchParams) {
+    const obj: Record<string, any> = {};
+    for (const [key, value] of body.entries()) {
+      obj[key] = value;
+    }
```
mrjvs, line 18: both formdata and urlsearch params can be turned into `[string, any][]` with just doing `[...obj]`.

Additionally, url search params can also be created from that in the constructor `new URLSearchParams([ ["key", "value"]])`. However for formdata you would need to do `const foo = new FormData; keys.forEach((entry) => foo.append(entry[0], entry[1])`;

either way, I think that structure may be more benificial instead of a record

mrjvs (changes requested):

src/backend/extension/request.ts
JorrinKievit: clean

mrjvs (approved):

PR #855, opened 2024-01-30
Fixed Sorting Shows

MemeCornucopia:
This pull request resolves Shows not updating in the Continue Watching List. The issue is let's say for example

I am on season three of Show B
I watch a new show called Show A
I go back to Show B and watch an episode or two of Show B
When I go to the home screen Show B is now on the top. 
On refresh Show B goes back to where it was before, and is not updated on the list.

### DIFF
--- a/src/backend/accounts/user.ts
+++ b/src/backend/accounts/user.ts
@@ -100,29 +100,22 @@ export function progressResponsesToEntries(responses: ProgressResponse[]) {
     }
 
     if (item.type === "show" && v.season.id && v.episode.id) {
-      if (
-        !item.seasons[v.season.id] ||
-        (item.episodes[v.episode.id] &&
-          new Date(v.updatedAt).getTime() >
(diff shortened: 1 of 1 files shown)


src/backend/accounts/user.ts
```diff
@@ -87,6 +87,11 @@ export function progressResponsesToEntries(responses: ProgressResponse[]) {
     }
 
     const item = items[v.tmdbId];
+    // Update the item only if the new update is more recent than the current one
```
binaryoverload, line 90:
```suggestion

    // Since each watched episode is a single array entry but with the same tmdbId, the root item updatedAt will only have the first episode's timestamp (which is not the newest).
    // Here, we are setting it explicitly so the updatedAt always has the highest updatedAt from the episodes.
```

binaryoverload (changes requested):

binaryoverload (approved):

PR #860, opened 2024-01-31
Copy playlist link

### LINKED ISSUE #857: [Feature]: Copy playlist link

Instead of downloading the HLS playlist, copy the download link.

Most HLS playlists don't have the base url prefixed in segment URLs
Playing directly from a HLS playlist's link works better

Assign me if approved 💓

ari-party:
This pull request resolves #857 

image
Allows left click and context menu > copy link address to copy.

### DIFF
--- a/src/assets/locales/en.json
+++ b/src/assets/locales/en.json
@@ -128,15 +128,11 @@
       },
       "morning": {
         "default": "What would you like to watch this morning?",
-        "extra": [
-          "I hear Before Sunrise is good"
-        ]
+        "extra": ["I hear Before Sunrise is good"]
       },
       "night": {
         "default": "What would you like to watch tonight?",
-        "extra": [
-          "Tired? I hear The Exorcist is good."
-        ]
+        "extra": ["Tired? I hear The Exorcist is good."]
       }
     }
   },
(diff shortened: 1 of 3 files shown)


zisra:
Using this would be more consistent with the rest of the codebase
```js
import { useCopyToClipboard } from "react-use";
```

src/assets/locales/en.json
```diff
@@ -243,7 +243,7 @@
     "menus": {
       "downloads": {
         "disclaimer": "Downloads are taken directly from the provider. movie-web does not have control over how the downloads are provided.",
-        "downloadPlaylist": "Download playlist",
+        "downloadPlaylist": "Copy playlist link",
```
mrjvs, line 246: You should change key from "downloadPlaylist" to "copyHlsPlaylist" and make the text "Copy HLS playlist link".
This is so the translations that currently exist for this key will fallback to english and not the old text.

mrjvs (changes requested):

mrjvs:
It seems you've missed the other request I had in the comment. There were two things addressed.

mrjvs (approved):

PR #874, opened 2024-02-03
fix(settings/appearance): Improved 'Active' pill

boidushya:
## Context

- `Active` pill under Settings > Appearance look out of the place right now since they are hardcoded to a purple-ish color. This PR fixes that by changing its background color based on whichever theme is selected. 
- This also provides yet another preview into what this selected theme might look like to the end user. 
- The padding and line-height was also adjusted to better represent a Pill. 

## Side by side comparison

| Theme | Before | After |
|---|--------|--------|
| Default | image | image |
| Blue | image | image |
| Teal | image | image |
| Red | image | image |
| Gray | image | image | 

### DIFF
--- a/src/pages/parts/settings/ThemePart.tsx
+++ b/src/pages/parts/settings/ThemePart.tsx
@@ -104,7 +104,7 @@ function ThemePreview(props: {
         <span className="font-medium text-white">{props.name}</span>
         <span
           className={classNames(
-            "inline-block px-3 text-sm transition-opacity duration-150 rounded-full bg-[#27182F] text-white",
+            "inline-block px-3 py-1 leading-tight text-sm transition-opacity duration-150 rounded-full bg-pill-background text-white/85 shadow-[rgba(255,255,255,0.15)_0_1px_0_inset,rgba(46,54,80,0.075)_0_1px_1px]",
             props.active ? "opacity-100" : "opacity-0 pointer-events-none",
           )}
         >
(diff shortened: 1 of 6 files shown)


src/pages/parts/settings/ThemePart.tsx
```diff
@@ -104,7 +104,7 @@ function ThemePreview(props: {
         <span className="font-medium text-white">{props.name}</span>
         <span
           className={classNames(
-            "inline-block px-3 text-sm transition-opacity duration-150 rounded-full bg-[#27182F] text-white",
+            "inline-block px-3 py-1 leading-tight text-sm transition-opacity duration-150 rounded-full bg-pill-background text-white/85 shadow-[rgba(255,255,255,0.15)_0_1px_0_inset,rgba(46,54,80,0.075)_0_1px_1px]",
```
mrjvs, line 107: Two small changes:
1. Instead of reusing the pill background. I suggest you make a new variable in the theme files. You can still use the same color, but it should get a different name since its a different purpose. You can put it under `themePreview.pill` or something similar. Theme files are in `/themes/list/*.ts` and `/themes/default.ts`
2. Please remove the inset shadow, it doesn't really fit with the movie-web brand.

mrjvs (changes requested):
I like the change, just two small nitpicks to address and its ready for merging.

src/pages/parts/settings/ThemePart.tsx
boidushya: Gotcha, will do!

boidushya:
image

Heres what it looks like after the recent commit 🤝

mrjvs (approved):
nice, thanks for the PR

PR #883, opened 2024-02-07
Add appearance preview

### LINKED ISSUE #868: [Feature]: Add appearance preview

Change the background colors when you change the appearance settings in order to know what it looks like without pressing "save" everytime.

This feature would be a lot more convenient to users when wanting to change their background colors. Having to choose a new appearance then clicking "save" to view what it looks like may be tedious if they do not like that color and decide to change it again.

_No response_

DylanBT928:
This pull request resolves #868


DylanBT928:
Not to sure what to do from here, but any help would be greatly appreciated.

mrjvs:
You could add an extra theme class on the body while you're on the settings page. It would overwrite the current theme. Then you wouldn't need to mess about with the theme store and it wouldn't automatically persist.

PR #886, opened 2024-02-08
Extend the length of captions

### LINKED ISSUE #870: [Feature]: PC - Extend the caption box length

Extend the caption box length so it is not bundled in the middle like this:
Bundled together

So it will be more like this:
longer

The mobile version works fine though. 

The subtitles are taking up too much space. 

Mobile works fine (from what I see at the moment

THX

Astro7145:
This pull request resolves #870 

I tried to extend the size of the box(where the captions are displayed) to make the captions longer, but I found that the size of the box doesn't do anything with making a new line. So, I thought the best way is to change the point when it makes a new line! (like, every 40 characters or 50 or something like that... and since you don't want the last word in a line to be separated in the middle of it when it makes the length more than 40 chars, I'm putting the word in the next line so it will look good!) Maybe, the way how I did it was wrong, but please let me know your appreciation and give me some advice to make it a better way!

### DIFF
--- a/src/components/player/base/SubtitleView.tsx
+++ b/src/components/player/base/SubtitleView.tsx
@@ -59,7 +59,7 @@ export function CaptionCue({
 
   return (
     <p
-      className="pointer-events-none mb-1 select-none rounded px-8 py-1 text-center leading-normal [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]"
+      className="pointer-events-none mb-1 select-none rounded px-4 py-1 text-center leading-normal [text-shadow:0_2px_4px_rgba(0,0,0,0.5)]"
       style={{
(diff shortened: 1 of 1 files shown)


binaryoverload:
Thank you for the PR. Unfortunately, after some internal discussion, we decided not to merge it.

This is separate from your code, which is very concise and well implemented, but instead, it has to do with the fact that this issue was approved under the assumption that movie-web was limiting the length of the captions.

Since it's actually not movie-web limiting the captions, and it's instead using the newlines from the subtitle file itself - we would like to keep it the way it is so that we remain faithful to the original subtitles - some of which use newlines for specific reasons.

Again, thanks for the PR, and we're sorry to waste your time.

PR #891, opened 2024-02-09
Fix HTTP error 403 on HLS stream after switching source

### LINKED ISSUE #887: [Bug]: hls: HTTP Error 403 after switching from an extension source
### What happened?

The player throws an hls: HTTP Error 403 on a Showbox stream after watching for a few minutes. This happens after switching from an HLS source (Vidplay).

Chrome

### Steps to reproduce?

1. Go to any show (I watched https://movie-web.app/media/tmdb-tv-37680-suits/98716/1540585)
2. Switch from Vidplay to Showbox
3. Watch for a few minutes.

It throws an error because it tries fetching the HLS stream even though I have already switched to the Showbox mp4 stream. It throws a 403 because this request has to go through the extension. In the screenshot below you can see the request to the Vidplay cdn.
image

JorrinKievit:
This pull request resolves #887 

I am not gonna lie I am unsure if this is the right approach. It seems to be doing an HLS request to the latest chunk minutes after removing the HLS instance. I have tried to use:
```
hls.stopLoad();
hls.removeAllListeners();
```
But that didn't change a thing. I have now chosen the approach of only emitting errors if the source that is failing, is the same as the one you are currently watching. This way it still does the failed request, but it won't show an error in the player.


mrjvs:
Honestly, it may be better if we ignore all fragment load errors. It's only been causing issues

JorrinKievit:
> Honestly, it may be better if we ignore all fragment load errors. It's only been causing issues

I guess the only issue you would have then is that when a source goes down mid watching, it will probably start infinitely loading without any feedback?

binaryoverload (approved):

PR #920, opened 2024-02-17
Add dynamic and engaging search bar placeholders

### LINKED ISSUE #918: [Feature]: Enhanced Search Bar Placeholder

To enhance user experience, I recommend adding an array of diverse placeholders to the search bar in our web app. By randomizing these placeholders with each user entry, we create a dynamic and engaging atmosphere. 

Users will encounter varied prompts like "Discover your perfect show or movie" or "Search for your preferred shows and movies" promoting a fresh and personalized interaction every time. This approach not only adds an element of surprise but also encourages users to approach their searches with renewed curiosity, contributing to a more positive and enjoyable web app experience.

_No response_

vijaysingh2219:
Enhance search bar experience with 5 dynamic placeholders for the movie web app. Provides a mix of engaging questions and suggestions, promoting user interaction and a personalized feel. Shuffled for a random appearance, contributing to a more dynamic user experience.

This pull request resolves #918 


src/assets/locales/en.json
```diff
@@ -116,7 +116,26 @@
       "failed": "Failed to find media, try again!",
       "loading": "Loading...",
       "noResults": "We couldn't find anything!",
-      "placeholder": "What do you want to watch?",
+      "placeholder": {
+        "default": "What do you want to watch?",
+        "extra": [
+          "What's on your movie radar today?",
+          "What's your cinematic preference today?",
```
JipFr, line 123: Am I crazy if I feel like this is just weird?

TheScreechingBagel: crazy? i was crazy once. they

src/assets/locales/en.json
```diff
@@ -116,7 +116,26 @@
       "failed": "Failed to find media, try again!",
       "loading": "Loading...",
       "noResults": "We couldn't find anything!",
-      "placeholder": "What do you want to watch?",
+      "placeholder": {
```
binaryoverload, line 119: I feel like this may cause an issue, as I don't know how Weblate will handle the change in type from string -> object

I think the best solution is to change the name of the key completely

binaryoverload (changes requested):

src/assets/locales/en.json
castdrian: what is it with the chatgpt responses

adlazaro:
Hey @vijaysingh2219, I really like the idea behind #918, but I feel like some of the current placeholder options might be too lengthy and verbose. The text could override the search bar on mobile devices:

movie-web-placeholder-long

binaryoverload (changes requested):
Mostly looks good, please just reduce the number of randomised English options to 5 - the current number is too many and will just mean the translators have the translate loads

binaryoverload:
In removing the excess from the list, please make sure to use the shorter ones so we don't run into the issue that @adlazaro talked about

binaryoverload (changes requested):
Thanks for the changes, please now fix the ESLint error on line 29 in `HeroPart.tsx`

binaryoverload (approved):

binaryoverload:
Looks good to me, I am going to ask @mrjvs to check I haven't missed anything

PR #938, opened 2024-02-23
Remove references to official domain

qtchaos:
This PR removes all links to the former official site and cancels all analytics related functions if backend is not set.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing Guidelines for movie-web
 
-Thank you for investing your time in contributing to our project! Your contribution will be reflected on [movie-web.app](https://movie-web.app).
+Thank you for investing your time in contributing to our project! Your contribution will be reflected on all of the community hosted instances that are on the latest version.
 
 Please read our [Code of Conduct](./CODE_OF_CONDUCT.md) to keep our community approachable and respectable.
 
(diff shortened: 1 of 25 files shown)


src/backend/accounts/progress.ts
```diff
@@ -103,6 +104,7 @@ export async function removeProgress(
   episodeId?: string,
   seasonId?: string,
 ) {
+  if (!BACKEND_URL) return;
```
binaryoverload, line 107: I think we should handle this higher up the chain because this doesn't account for users overriding the backend URL. In `ProgessSyncer.tsx` we call the `useBackendUrl` which could return undefined.

Could you modify the return type of `useBackendUrl` to return `string | undefined`, and then modify `ProgessSyncer.tsx` to return early if the URL is undefined

binaryoverload (changes requested):

binaryoverload (approved):

.github/CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ Here are some tips to make sure that your pull requests are :pinched_fingers: fi
 ### Language Contributions
 Language contributions help movie-web massively, allowing people worldwide to use our app!
 
-We use weblate for crowdsourcing our translations. [Click here to go to our translation tool.](https://weblate.movie-web.app/projects/movie-web/website/)
+We use weblate for crowdsourcing our translations.
 
 1. First make sure you make an account. (click the link above)
```
TheScreechingBagel, line 90: there's no new "link above" yet btw

qtchaos: ~~forgot about that one~~ just as planned... ill add it back once we get a pretty domain for weblate 😎

PR #939, opened 2024-02-25
Update the admin page to run workers in parallel

qtchaos:
Changes the admin page to run all worker tests at the same time, allowing for a faster test. Since I fear that people would spam this, I also added a 5 second cooldown period after each test, which disables the button.

Also changed a bit of the TMDB tests grammar.

### DIFF
--- a/src/pages/parts/admin/WorkerTestPart.tsx
+++ b/src/pages/parts/admin/WorkerTestPart.tsx
@@ -52,14 +52,18 @@ export function WorkerTestPart() {
     { id: string; status: "error" | "success"; error?: Error }[]
   >([]);
 
+  const [buttonDisabled, setButtonDisabled] = useState(false);
+
   const [testState, runTests] = useAsyncFn(async () => {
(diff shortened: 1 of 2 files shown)


src/pages/parts/admin/WorkerTestPart.tsx
```diff
@@ -85,7 +89,10 @@ export function WorkerTestPart() {
           error: err as Error,
         });
       }
-    }
+    });
+
+    await Promise.all(workerPromises);
+    setTimeout(() => setButtonDisabled(false), 5000);
```
JorrinKievit, line 95: You might have to cleanup the setTimeout function in a return (I could be wrong, but I thought that was good practice in useEffect, but idk the behaviour of useAsyncFn)

qtchaos: Ah yeah, you might be right

binaryoverload (approved):

PR #946, opened 2024-02-27
Update Dockerfile and add compose.yaml file 

Audiosutras:
This pull request resolves the absence of environment variables from the Dockerfile and adds a compose.yaml file based on a discussion between Blade and BinaryOverload.


Dockerfile
```diff
@@ -9,8 +9,36 @@ COPY pnpm-lock.yaml ./
...
+ARG CORS_PROXY_URL=""
+ARG DMCA_EMAIL=""
+ARG NORMAL_ROUTER="false"
+ARG BACKEND_URL=""
+ARG HAS_ONBOARDING="false"
+ARG ONBOARDING_CHROME_EXTENSION_INSTALL_LINK=""
+ARG ONBOARDING_PROXY_INSTALL_LINK=""
+ARG DISALLOWED_IDS=""
+ARG CDN_REPLACEMENTS=""
+ARG TURNSTILE_KEY=""
```
binaryoverload, line 25: Instead of using `=""` to default to an empty list, you can just do `ARG GA_ID` for example

compose.yaml
```diff
@@ -0,0 +1,27 @@
+version: "3.8"
+
+services:
+
+  movieweb:
+    image: ghcr.io/movie-web/movie-web:latest
```
binaryoverload, line 6: Instead of using the image, you should use `build` as you can't specify environment variables at runtime

The build field can be specified with args as follows:

```yml
build:
  context: .
  args:
    GA_ID: example
```

compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+    ports:
+      - "3000:80"
+    restart: always
+    environment:
```
binaryoverload, line 10: These environment variables should be build time args as demoed in my previous comment

compose.yaml
binaryoverload, line 1: This file should be named `docker-compose.yml`

binaryoverload (changes requested):
Looks good for the most part, just some tweaks to the Docker compose file.

Thanks!

Dockerfile
Audiosutras: Great to know. Changed

compose.yaml
Audiosutras: I see, I got turned around from the conversation earlier. I thought this compose file was going to be a standalone reference for people to just copy it and run the project.

Audiosutras: changed.

compose.yaml
Audiosutras: removed

compose.yaml
Audiosutras: done.

compose.yaml
binaryoverload: > I see, I got turned around from the conversation earlier. I thought this compose file was going to be a standalone reference for people to just copy it and run the project.

Because of the way we build the project, the way it has to be done is that people can clone the repo and run docker-compose from the repo with the correct build args

Audiosutras:
> Looks good for the most part, just some tweaks to the Docker compose file.
> 
> Thanks!

I also changed port `3000:80` back to `80:80`

compose.yaml
binaryoverload: The other alternative is mounting a `config.js` file in the Docker container

docker-compose.yaml
```diff
@@ -0,0 +1,10 @@
+version: "3.8"
+
+services:
+
+  movieweb:
+    build:
+      context: .
+    ports:
+      - "80:80"
+    restart: always
```
DenuxPlays, line 10:
```suggestion
    restart: unless-stopped
```
we do not want that the container starts after we stopped it

binaryoverload: Good spot, thanks @DenuxPlays

compose.yaml
Audiosutras: Gotcha. Thanks for the explanation.

binaryoverload (approved):

PR #947, opened 2024-02-27
Extension: Prepare stream for all domains in a hls playlist

### LINKED ISSUE #905: [Bug]: Extension does not work for Closeload
### What happened?

The Closeload embed does not work with the extension. This is because there are different domains in the hls playlist files. Only the main domain is prepared through the extension. As you can see in the screenshot, a domain rule is set for hls.closeload, but not for  srv9.servercdn1035

_No response_

### Steps to reproduce?

1. Use extension v1.1.0
2. Choose a movie, and Ridomovies

image
image
image
image

JorrinKievit:
This pull request resolves #905 and VidSrcTo

There was also a bug in the hlsLevelQuality function, sometimes levels don't exist and it throws an error which causes the code after not to run.

### DIFF
--- a/src/components/player/display/base.ts
+++ b/src/components/player/display/base.ts
@@ -152,10 +152,9 @@ export function makeVideoElementDisplayInterface(): DisplayInterface {
 
           if (isExtensionActiveCached()) {
             hls.on(Hls.Events.LEVEL_LOADED, async (_, data) => {
-              console.log(data);
-              const chunkUrlsDomains = data.details.fragments
-                .map((v) => v.url)
-                .map((v) => new URL(v).hostname);
+              const chunkUrlsDomains = data.details.fragments.map(
+                (v) => new URL(v.url).hostname,
+              );
               const chunkUrls = [...new Set(chunkUrlsDomains)];
 
               await setDomainRule({
--- a/src/components/player/display/base.ts
+++ b/src/components/player/display/base.ts
@@ -2,6 +2,11 @@ import fscreen from "fscreen";
 import Hls, { Level } from "hls.js";
 
 import {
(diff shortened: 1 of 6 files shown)


sudo-flix:
looks fire

src/components/player/display/base.ts
```diff
@@ -144,6 +149,24 @@ export function makeVideoElementDisplayInterface(): DisplayInterface {
           if (!hls) return;
           reportLevels();
           setupQualityForHls();
+
+          if (isExtensionActiveCached()) {
+            hls.on(Hls.Events.LEVEL_LOADED, async (_, data) => {
+              const chunkUrlsDomains = data.details.fragments.map(
+                (v) => new URL(v.url).hostname,
```
binaryoverload, line 156: Is there any chance of a URL being undefined/invalid format on a fragment - worth handling?

binaryoverload (changes requested):

src/components/player/display/base.ts
JorrinKievit: Good point, not sure if hls.js handles that. Will check when I get the time (or if someone else wants to 😉)

qtchaos: Would something like this be enough? If the URL is null we just use an empty string which will make the request 403 (black screen on player)

```ts
const chunkUrlsDomains = data.details.fragments.map(
  (v) => new URL(v.url).hostname ?? "",
);
```

JorrinKievit: Issue is more that new URL() can throw an error if the input is invalid

qtchaos (approved):

PR #958, opened 2024-02-29
Automatically prefix backend url with https:// if not provided

### LINKED ISSUE #956: [Feature]: Correct Backend URLs to Include https:// Prefix

Using a backend URL without the https:// prefix causes an error when attempting to log in:
`[POST] "backend.example.com/auth/login/start": 404`
It makes sense that this would happen, but it would be better if backend URLs were simply corrected to include https:// if the user doesn’t type this prefix in. 

This would reduce confusion for users and new selfhosters attempting to use a backend for the first time

_No response_

binaryoverload:
This pull request resolves #956 

### DIFF
--- a/src/pages/Settings.tsx
+++ b/src/pages/Settings.tsx
@@ -208,7 +208,13 @@ export function SettingsPage() {
     // when backend url gets changed, log the user out first
     if (state.backendUrl.changed) {
       await logout();
-      setBackendUrl(state.backendUrl.state);
+
+      let url = state.backendUrl.state;
+      if (url && !url.startsWith("http://") && !url.startsWith("https://")) {
+        url = `https://${url}`;
+      }
+
+      setBackendUrl(url);
     }
   }, [
     state,


src/pages/Settings.tsx
```diff
@@ -208,7 +208,13 @@ export function SettingsPage() {
     // when backend url gets changed, log the user out first
     if (state.backendUrl.changed) {
       await logout();
-      setBackendUrl(state.backendUrl.state);
+
+      let url = state.backendUrl.state;
+      if (url && !url.startsWith("http://") && !url.startsWith("https://")) {
+        url = `https://${url}`;
+      }
```
mrjvs, line 215: what about people using `postgresql://` protocol for their backend?

JorrinKievit (approved):

src/pages/Settings.tsx
binaryoverload: Well, then they are silly

PR #961, opened 2024-03-01
Add vercel config to properly support non-hash router

binaryoverload:


TheScreechingBagel:
does that actually make it work?

binaryoverload:
> does that actually make it work?

I have tested it, it should work.

JorrinKievit (commented):
Should we even include this by default in this repo, even though some do not host it on vercel?

Shouldnt we just document this file in the docs instead?

binaryoverload:
My thinking is that it doesn't (or shouldn't) harm anything by being in here

By including it in the repo, we don't need an extra docs step for people to follow and it also means they don't have to edit their repo which would make it out of sync with the fork

JorrinKievit (approved):
Fair enough

PR #998, opened 2024-03-11
Fix Nan% in Background Blur

Honkertonken:
This pull request resolves #XXX

Fixes the bugged Nan%

This issue only appeared on firefox on pc (wasnt a problem on edge/chrome)
Havent tested on mobile

Before : 
image

After
image


binaryoverload:
Does this fix the issue? 

I'm not convinced this would as calling `.toFixed` on a NaN results in `NaN`

Honkertonken:
Redeployed (vercel) and tested and can confirm this isnt needed.

The Nan% still appears when using the "default" link but it works just fine if you use the "deployment" or "branch" link 
Not quite sure what the problem is.

Apologies for the inconvenience, should have taken a better look.

6uArwXQCs
6uArEpw3z

PR #1007, opened 2024-03-12
Add embedded subtitles support

### LINKED ISSUE #906: [Bug]: Unable to switch subtitles on Ridomovies
### What happened?

The Ridoo embed has the subtitles/captions in the HLS playlist. The first subtitle is shown in the player, but the subtitle menu does not show the the other subtitles that are in this HLS playlist. This means the Chinese one is always shown with no option of selecting another language.

_No response_

### Steps to reproduce?

1. Go to https://movie-web.app/media/tmdb-tv-84958-loki/114355/2534997
2. Use Ridomovies

Below is an example of a master HLS playlist file with subtitles:
image
image

ssssobek:
This pull request resolves #906

 - [ ] I have tested all of my changes. - I've tested it on almost every episode from #906, but I reckon it needs testing in other cases (I couldn't find any)

# Issues
- Loading the subtitles can be quite slow, probably because there is like 300 vtt files to fetch.
- Subtitle language isn't persistent when switching to source with embedded subtitles (probably because the settings load faster than hls)

# Notes
- There probably are a few debug prints, will be removed when it's certain that this thing works.

# Preview (on media from #906)
https://github.com/movie-web/movie-web/assets/95639019/eae27deb-ac84-49d3-b06f-6ccf549f58b4

### DIFF
--- a/src/components/player/hooks/useCaptions.ts
+++ b/src/components/player/hooks/useCaptions.ts
@@ -1,6 +1,8 @@
-import { useCallback } from "react";
+import { useCallback, useMemo } from "react";
+import subsrt from "subsrt-ts";
+import { ContentCaption } from "subsrt-ts/dist/types/handler";
 
-import { downloadCaption } from "@/backend/helpers/subs";
+import { downloadCaption, downloadWebVTT } from "@/backend/helpers/subs";
 import { usePlayerStore } from "@/stores/player/store";
 import { useSubtitleStore } from "@/stores/subtitles";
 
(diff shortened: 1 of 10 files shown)


binaryoverload:
This seems to work well, thanks for your work! One of us will review the code properly soon :D

JorrinKievit:
Heya, I just did a quick test too. It seems to error out on all languages except for the one that is initially loaded (for me that was Chinese).

JorrinKievit:
Aha, loaded into another browser for the first time, and I can reproduce it again

JorrinKievit:
Wow, nevermind. In the end I tracked it down to hls?.setSubtitleOption not existing as a method even though I reinstalled hls with the new version. After removing vite cache everything works 🎉

ssssobek:
Glad it turned out fine, lol

src/components/player/hooks/useCaptions.ts
```diff
@@ -12,32 +14,106 @@ export function useCaptions() {
...
+        const fragments =
+          track.details?.fragments?.filter(
+            (frag) => frag !== null && frag.url !== null,
+          ) ?? [];
+
+        const vttCaptions = (
+          await Promise.all(
+            fragments.map(async (frag) => {
+              const vtt = await downloadWebVTT(frag.url);
+              const parsed = subsrt.parse(vtt);
```
JorrinKievit, line 63: You could use return parseSubtitles(vtt) from /player/utils/captions.ts here

src/components/player/hooks/useCaptions.ts
```diff
@@ -12,32 +14,106 @@ export function useCaptions() {
...
+              return parsed.filter(
+                (c) => c.type === "caption",
+              ) as ContentCaption[];
+            }),
+          )
+        ).flat();
+
+        // for some reason, in some cases there will be captions
+        // with the same start/end times, the same text duplicated
+        const filtered = vttCaptions.reduce(
```
JorrinKievit, line 73: Could we extract this logic to a filterDuplicateCaptionCues function in captions.ts file too? Makes this code a bit easier to follow

JorrinKievit (changes requested):
I think its fine to remove the console.logs. Everything seems to work well!

src/components/player/hooks/useCaptions.ts
```diff
@@ -12,32 +14,106 @@ export function useCaptions() {
...
-        id: caption.id,
-        language: caption.language,
-        srtData,
-        url: caption.url,
-      });
-      resetSubtitleSpecificSettings();
-      setLanguage(caption.language);
+      if (!caption.hls) {
+        const srtData = await downloadCaption(caption);
+        setCaption({
```
JorrinKievit, line 38: These 3 methods are called in both conditions. Let's make a local copy of this object:

```
const captionToSet: Caption = {
        id: caption.id,
        language: caption.language,
        url: caption.url,
        srtData: "",
};
```
Set the srtData in both conditions, and call setCaption, resetSubtitleSpecificSettings and setLanguage at the end

JorrinKievit (changes requested):

src/components/player/hooks/useCaptions.ts
```diff
@@ -1,9 +1,27 @@
...
 
-import { downloadCaption } from "@/backend/helpers/subs";
+import { downloadCaption, downloadWebVTT } from "@/backend/helpers/subs";
+import { Caption } from "@/stores/player/slices/source";
 import { usePlayerStore } from "@/stores/player/store";
 import { useSubtitleStore } from "@/stores/subtitles";
 
+import { parseVttSubtitles } from "../utils/captions";
+
+const filterDuplicateCaptionCues = (cues: ContentCaption[]) =>
```
JorrinKievit, line 12: Could you move this to /player/utils/captions.ts instead? This file is for react hooks, and that file is more for utility functions for captions

JorrinKievit (changes requested):

JorrinKievit:
After the last fix, it's ready to be merged! ❤️

JorrinKievit (approved):

Issue #900, opened 2024-02-11
[Feature]: 


luvsense:
### What feature do you want to add?

Direct Bookmarking from Movie Thumbnails

### Why do you want to have this feature?

I want to add this feature to make bookmarking movies more straightforward and efficient. By enabling users to bookmark directly from the movie thumbnails, it eliminates the need to open each movie to bookmark it.

PR #1044, opened 2024-03-27
Filter unreleased shows/movies. 

### LINKED ISSUE #357: [Feature]: Filter out unreleased movies on search

Movies that are not yet released should not be included in the search result.

This site is not a movie info site like imdb, but an actual site to watch movies. Nobody cares about unreleased stuff.

_No response_

Honkertonken:
This pull request resolves #357 

image
image


binaryoverload (changes requested):
Thanks for the PR. Unfortunately, this doesn't quite line up with our intentions

We want to make it so that unreleased shows are still shown, just not clickable. This will mean that there's no confusion about the movie/show not existing.

If you want to do #900 at the same time that would be perfect for this, but is not a blocker for this PR.

To show that something is unreleased, I think we should add opacity to the card and have the label of "Unreleased" where the year would normally be

Honkertonken:
Made the necessary changes 

Preview - 

img

src/components/media/MediaCard.tsx
```diff
@@ -35,10 +35,14 @@ function MediaCardContent({
...
 
-  const canLink = linkable && !closable;
+  const canLink = linkable && !closable && !!media.year;
 
   const dotListContent = [t(`media.types.${media.type}`)];
-  if (media.year) dotListContent.push(media.year.toFixed());
+  if (media.year) {
+    dotListContent.push(media.year.toFixed());
+  } else {
+    dotListContent.push(t("Unreleased"));
```
binaryoverload, line 44: Please make this an actual language string - we use paths for the key instead of the original lang text

`media.unreleased` would be good

src/components/media/MediaCard.tsx
```diff
@@ -35,10 +35,14 @@ function MediaCardContent({
   const { t } = useTranslation();
   const percentageString = `${Math.round(percentage ?? 0).toFixed(0)}%`;
 
-  const canLink = linkable && !closable;
+  const canLink = linkable && !closable && !!media.year;
 
   const dotListContent = [t(`media.types.${media.type}`)];
-  if (media.year) dotListContent.push(media.year.toFixed());
+  if (media.year) {
```
binaryoverload, line 41: Instead of checking only if the year is present, we should also check if it is greater than the current year. 

i.e. if a media has a year of 2025, that should also not be linkable and should have "Unreleased" added to the dot list

src/components/media/MediaCard.tsx
```diff
@@ -157,7 +161,7 @@ export function MediaCard(props: MediaCardProps) {
     }
   }
 
-  if (!props.linkable) return <span>{content}</span>;
+  if (!props.linkable || !props.media.year) return <span>{content}</span>;
```
binaryoverload, line 164: Please alter the year check as described before to check for years in the future

binaryoverload (changes requested):
Thanks for making the requested changes - just a couple of comments and then this is good to go!

Honkertonken:
Should be good to go 

Preview 
img1
img2

binaryoverload (changes requested):
Looking good!

Could we have it so that an unreleased media that has a year shows it?

"Movie . 2020 . Unreleased" _should_ fit

src/components/media/MediaCard.tsx
```diff
@@ -35,10 +35,16 @@ function MediaCardContent({
   const { t } = useTranslation();
   const percentageString = `${Math.round(percentage ?? 0).toFixed(0)}%`;
 
-  const canLink = linkable && !closable;
+  const currentYear = new Date().getFullYear();
+  const canLink =
+    linkable && !closable && media.year && media.year < currentYear;
```
JorrinKievit, line 40: I dont think this works, because this would show only movies/shows from the year 2023 and earlier 😁

JorrinKievit (changes requested):

Honkertonken:
Done
Preview 
im1
im2
im3

binaryoverload (approved):
Looks good, thanks for being patient with the changes!

@JorrinKievit could you have a 2nd glance before I merge? Thanks :)

jacquesvfd:
Are we sure we don't want unreleased movies to be clickable? Occasionally movies are available before their release date, so it might be better to let the user try anyway. Plus, I imagine having different release dates in different regions creates more potential to break otherwise working movies. 
It might be better to grey them out but leave them clickable, and perhaps add a warning dialogue to tell the user that it's unlikely to be available because it's not released yet.

zisra:
>  Occasionally movies are available before their release date, so it might be better to let the user try anyway
I second this.

JipFr:
The same issue was raised for the episode selector without any evidence whatsoever. When are movies and episodes ever available before their release date?

The only situation in which I can imagine this happening is the release date being the date on which it comes to streaming, DVD, or Blu-Ray, but that doesn't seem like a particularly big concern.

jacquesvfd:
> When are movies and episodes ever available before their release date?

I watched one of the Trolls sequels a few days before release several years ago. It happens. I like having it marked to tell users it's not out yet, but there it doesn't hurt to let the user _try_ if they want to.

blikje7up:
@JipFr It has happened a few times, like with game of thrones (actually happend multiple times).
If someone has idea's for a clean implementation on how to notify the user but also let them click through it would be nice but definitely not necessary.

JipFr:
I feel, and this is just me, that this is an issue so incredibly rare that it's not worth adding to the site. If somebody _knows_ something has been leaked and they want to see it, let them torrent it. They're not our average user.

TheScreechingBagel:
another reason to have them still be clickable: now people aren’t able to open the unreleased piece of media to bookmark

JorrinKievit:
Which is why this exist https://github.com/movie-web/movie-web/issues/900

PR #1098, opened 2024-04-08
Fixed next episode button not loading new season

### LINKED ISSUE #1049: [Feature] - Show a "Next Season" button when at the end of a TV season
### What happened?

when at the last episode in the season, even if there is another season there is no NextEpisodeButton because of his the nextEp var is set

Microsoft Edge

### Steps to reproduce?

1. go to a show and click on it
2. go to the last episode of the season
3. fast forward to ~10 seconds before episode depletion 
4. don’t see a button for the next episode even if there’s another season

_No response_

dro-1:
This pull request resolves #1049.
The next episode button was not showing when the last episode of a season was reached. It is now showing and loads said episode when clicked.

### DIFF
--- a/src/components/player/atoms/NextEpisodeButton.tsx
+++ b/src/components/player/atoms/NextEpisodeButton.tsx
@@ -1,13 +1,18 @@
 import classNames from "classnames";
 import { useCallback } from "react";
 import { useTranslation } from "react-i18next";
+import { useAsync } from "react-use";
 
+import { getMetaFromId } from "@/backend/metadata/getmeta";
+import { MWMediaType, MWSeasonMeta } from "@/backend/metadata/types/mw";
 import { Icon, Icons } from "@/components/Icon";
 import { usePlayerMeta } from "@/components/player/hooks/usePlayerMeta";
 import { Transition } from "@/components/utils/Transition";
 import { PlayerMeta } from "@/stores/player/slices/source";
 import { usePlayerStore } from "@/stores/player/store";
 import { useProgressStore } from "@/stores/progress";
+
+import { hasAired } from "../utils/aired";
 
 function shouldShowNextEpisodeButton(
   time: number,
@@ -39,6 +44,45 @@ function Button(props: {
   );
 }
 
+function useSeasons(mediaId: string, isLastEpisode: boolean = false) {
+  const state = useAsync(async () => {
+    if (isLastEpisode) {
+      const data = await getMetaFromId(MWMediaType.SERIES, mediaId ?? "");
+      if (data?.meta.type !== MWMediaType.SERIES) return null;
+      return data.meta.seasons;
+    }
+  }, [mediaId, isLastEpisode]);
+
+  return state;
+}
+
+function useNextSeasonEpisode(
+  nextSeason: MWSeasonMeta | undefined,
+  mediaId: string,
+) {
+  const state = useAsync(async () => {
+    if (nextSeason) {
+      const data = await getMetaFromId(
+        MWMediaType.SERIES,
+        mediaId ?? "",
+        nextSeason?.id,
+      );
+      if (data?.meta.type !== MWMediaType.SERIES) return null;
+
+      const nextSeasonEpisodes = data?.meta?.seasonData?.episodes
+        .filter((episode) => hasAired(episode.air_date))
+        .map((episode) => ({
+          number: episode.number,
+          title: episode.title,
+          tmdbId: episode.id,
+        }));
+
+      if (nextSeasonEpisodes.length > 0) return nextSeasonEpisodes[0];
+    }
+  }, [mediaId, nextSeason?.id]);
+  return state;
+}
+
 export function NextEpisodeButton(props: {
   controlsShowing: boolean;
   onChange?: (meta: PlayerMeta) => void;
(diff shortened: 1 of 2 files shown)


src/components/player/atoms/NextEpisodeButton.tsx
```diff
@@ -47,17 +52,61 @@ export function NextEpisodeButton(props: {
...
+    meta?.tmdbId ?? "",
+    meta?.season?.tmdbId ?? "",
+  );
   const { setDirectMeta } = usePlayerMeta();
   const hideNextEpisodeButton = usePlayerStore((s) => s.hideNextEpisodeButton);
   const metaType = usePlayerStore((s) => s.meta?.type);
   const time = usePlayerStore((s) => s.progress.time);
   const showingState = shouldShowNextEpisodeButton(time, duration);
   const status = usePlayerStore((s) => s.status);
+  const [nextSeason, setNextSeason] = useState<MWSeasonMeta>();
```
JorrinKievit, line 65: I dont think we need this state. a simple local variable 
 ```
const nextSeason = seasons?.find(
    (season) => season.number === (meta?.season?.number ?? 0) + 1,
  );
```

would suffice

File: src/components/player/atoms/NextEpisodeButton.tsx [hunk 2]
```diff
@@ -47,17 +52,61 @@ export function NextEpisodeButton(props: {
   const duration = usePlayerStore((s) => s.progress.duration);
   const isHidden = usePlayerStore((s) => s.interface.hideNextEpisodeBtn);
   const meta = usePlayerStore((s) => s.meta);
+  const [_, seasons] = useSeasonData(
```
JorrinKievit, line 55: This does not seem necessary if the current episode is not the last episode. Can we make sure this is only called when that is the case? This way we can prevent an unneeded call

src/components/player/atoms/NextEpisodeButton.tsx
```diff
@@ -47,17 +52,61 @@ export function NextEpisodeButton(props: {
...
+  useEffect(() => {
+    if (seasons) {
+      const next = seasons.find(
+        (season) => season.number === (meta?.season?.number ?? 0) + 1,
+      );
+      setNextSeason(next);
+    }
+  }, [meta?.season?.number, seasons]);
+
+  useEffect(() => {
```
JorrinKievit, line 87: We could use the useAsync hook here. Then we can also get rid of the `nextSeasonEpisodes` local state. We also do not need all the episodes, so only returning the first one from the useAsync hook would suffice.

src/components/player/atoms/NextEpisodeButton.tsx
```diff
@@ -47,17 +52,61 @@ export function NextEpisodeButton(props: {
...
+      title: string;
+      tmdbId: string;
+    }[]
+  >([]);
   const setShouldStartFromBeginning = usePlayerStore(
     (s) => s.setShouldStartFromBeginning,
   );
   const updateItem = useProgressStore((s) => s.updateItem);
 
+  useEffect(() => {
```
JorrinKievit, line 78: We can remove this when we use the local nextSeason variable

JorrinKievit (changes requested):

JorrinKievit (dismissed):
Also the button should show "Next season" when its on the last episode, and there is a next season

dro-1:
Made the requested changes @JorrinKievit . 
Awaiting review.

src/components/player/atoms/NextEpisodeButton.tsx
```diff
@@ -121,7 +190,10 @@ export function NextEpisodeButton(props: {
           className="bg-buttons-primary hover:bg-buttons-primaryHover text-buttons-primaryText flex justify-center items-center"
         >
           <Icon className="text-xl mr-1" icon={Icons.SKIP_EPISODE} />
-          {t("player.nextEpisode.next")}
+          {isLastEpisode && nextEp
+            ? t("player.nextEpisode.nextSeason")
+            : t("player.nextEpisode.next")}
+          {}
```
qtchaos, line 196: stray braces?

qtchaos (changes requested):

src/components/player/atoms/NextEpisodeButton.tsx
```diff
@@ -39,6 +44,45 @@ function Button(props: {
...
+
+      const nextSeasonEpisodes = data?.meta?.seasonData?.episodes
+        .filter((episode) => hasAired(episode.air_date))
+        .map((episode) => ({
+          number: episode.number,
+          title: episode.title,
+          tmdbId: episode.id,
+        }));
+
+      if (nextSeasonEpisodes.length) return nextSeasonEpisodes[0];
```
binaryoverload, line 80: I generally dislike falsy checks, they are pretty unclear when scanning code

```suggestion
      if (nextSeasonEpisodes.length > 0) return nextSeasonEpisodes[0];
```

binaryoverload (changes requested):

src/components/player/atoms/NextEpisodeButton.tsx
dro-1: oh yeah, they definitely are. I'll fix them right away.
do you think it might be reasonable to add that rule to the linter so we catch it next time?

qtchaos: If there's a rule for that then I wouldn't be against it

dro-1:
Fixed the issues and squashed the commits into one.
I didn't find any linter rule for empty braces so I left that.

binaryoverload:
@dro-1 FYI, generally, it's bad practice to rebase while your PR is open for review, as it makes it harder for reviewers to see what you've done 

I'll test the PR then LGTM

binaryoverload (approved):

dro-1:
sorry about that @binaryoverload , I'll avoid it in future PRs.
what does LGTM mean 👀

qtchaos:
> what does LGTM mean 👀

Looking good to me

PR #1108, opened 2024-04-12
Added alternate tmdb endpoint to fallback when main url is blocked

### LINKED ISSUE #1107: [Feature]: Provide ability to use alternate TMDB api incase of failure using the default api

If the search fails through the main TMDB API endpoint, then we can try having the search redirect to the alternate endpoint, or we can have an optional environment variable that can be used to manage these endpoints that the self-hoster can change as per their target audience. 

In India and maybe elsewhere, major internet providers like Jio block the main TMDB URL, [https://api.themoviedb.org](https://api.themoviedb.org), however, the mirror, [https://api.tmdb.org](https://api.tmdb.org), is not blocked. 

We can additionally allow users to use this simple proxy [tmdb-proxy](https://github.com/chervontsev/tmdb-proxy) to self-host their own endpoint to circumvent this restriction or use the tmdb mirror endpoint for the solution. I can work on this and create a pull request if required for fixing this issue. A lot of such issues have been closed without any resolution, and this can help a lot of self-hosting users.

meghrathod:
This pull request resolves #1107 

### DIFF
--- a/src/backend/metadata/tmdb.ts
+++ b/src/backend/metadata/tmdb.ts
@@ -143,8 +143,9 @@ export function decodeTMDBId(
   };
 }
 
+const baseURL = "https://api.themoviedb.org/3";
 const otherUrl = "https://api.tmdb.org/3";
-let baseURL = "https://api.themoviedb.org/3";
+let useFallback = false;
 
 const apiKey = conf().TMDB_READ_API_KEY;
 
@@ -155,33 +156,26 @@ const headers = {
 
 async function get<T>(url: string, params?: object): Promise<T> {
   if (!apiKey) throw new Error("TMDB API key not set");
-
   let res: T;
-
   try {
     res = await mwFetch<T>(encodeURI(url), {
       headers,
-      baseURL,
+      baseURL: !useFallback ? baseURL : otherUrl,
       params: {
         ...params,
       },
-      signal: AbortSignal.timeout(baseURL !== otherUrl ? 5000 : 30000),
+      signal: AbortSignal.timeout(!useFallback ? 5000 : 30000),
     });
   } catch (err) {
-    if (baseURL !== otherUrl) {
-      baseURL = otherUrl;
-      res = await mwFetch<T>(encodeURI(url), {
-        headers,
-        baseURL,
-        params: {
-          ...params,
-        },
-      });
-    } else {
-      res = Promise.reject();
-    }
+    useFallback = true;
+    res = await mwFetch<T>(encodeURI(url), {
+      headers,
+      baseURL: otherUrl,
+      params: {
+        ...params,
+      },
+    });
   }
-
   return res;
 }
 
--- a/src/backend/metadata/tmdb.ts
+++ b/src/backend/metadata/tmdb.ts
@@ -156,10 +156,10 @@ const headers = {
 async function get<T>(url: string, params?: object): Promise<T> {
   if (!apiKey) throw new Error("TMDB API key not set");
 
-  let res: Promise<T>;
+  let res: T;
 
   try {
-    res = await mwFetch<any>(encodeURI(url), {
+    res = await mwFetch<T>(encodeURI(url), {
       headers,
       baseURL,
       params: {
@@ -170,7 +170,7 @@ async function get<T>(url: string, params?: object): Promise<T> {
   } catch (err) {
     if (baseURL !== otherUrl) {
       baseURL = otherUrl;
-      res = await mwFetch<any>(encodeURI(url), {
+      res = await mwFetch<T>(encodeURI(url), {
         headers,
         baseURL,
         params: {
--- a/src/backend/metadata/tmdb.ts
+++ b/src/backend/metadata/tmdb.ts
@@ -156,13 +156,6 @@ const headers = {
 async function get<T>(url: string, params?: object): Promise<T> {
   if (!apiKey) throw new Error("TMDB API key not set");
 
-  const controller = new AbortController();
-  const { signal } = controller;
-
-  const timeoutId =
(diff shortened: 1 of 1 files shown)


src/backend/metadata/tmdb.ts
```diff
@@ -155,13 +156,40 @@ const headers = {
...
-    params: {
-      ...params,
-    },
-  });
+  const controller = new AbortController();
+  const { signal } = controller;
+
+  const timeoutId =
+    baseURL === otherUrl
+      ? setTimeout(() => controller.abort(), 15000)
```
JorrinKievit, line 164: Why do we need to work with timeouts, and not just use ` { signal: AbortSignal.timeout(5000)` ?

JorrinKievit (changes requested):
I have some thoughts on this one. Because for people where tmdb is blocked, this would mean that every request to tmdb will take 3+ seconds. Even though after the first try we pretty much know that the requests in the future will also fail. 

Maybe we can have an approach similarly to how we implemented `isExtensionActiveCached` ?

1. Do a request, if it fails but the second one works set a boolean that it should use the fallback one.
2. For all tmdb requests after, use the fallback URL.

I think its worth exploring different implementations here, because the current UX could be better

src/backend/metadata/tmdb.ts
meghrathod: Sure, we can do this. Didn't look at this.

meghrathod:
This is what will happen in my current implementation from the second request onwards all request are served by the new URL as it will become the base URL. I still feel an environment variable must also be present if both the URLs are blocked. Happy know your thoughts.

src/backend/metadata/tmdb.ts
meghrathod: Made this change, pls review. Excuse me as I'm starting out in opensource dev so might make some mistakes.

src/backend/metadata/tmdb.ts
```diff
@@ -143,7 +143,8 @@ export function decodeTMDBId(
   };
 }
 
-const baseURL = "https://api.themoviedb.org/3";
+const otherUrl = "https://api.tmdb.org/3";
+let baseURL = "https://api.themoviedb.org/3";
```
JorrinKievit, line 147: In my opinion this would be more readable and understandable if we keep track with a shouldUseFallback boolean, and use one of the URLs based on that.

JorrinKievit (changes requested):

src/backend/metadata/tmdb.ts
```diff
@@ -155,13 +156,32 @@ const headers = {
...
+    res = await mwFetch<any>(encodeURI(url), {
+      headers,
+      baseURL,
+      params: {
+        ...params,
+      },
+      signal: AbortSignal.timeout(baseURL !== otherUrl ? 5000 : 30000),
+    });
+  } catch (err) {
+    if (baseURL !== otherUrl) {
```
binaryoverload, line 171: Why are we doing checking if the `baseURL` equals the `otherUrl`? This condition will always be false as they are constants

src/backend/metadata/tmdb.ts
```diff
@@ -155,13 +156,32 @@ const headers = {
...
-    headers,
-    baseURL,
-    params: {
-      ...params,
-    },
-  });
+  let res: Promise<T>;
+
+  try {
+    res = await mwFetch<any>(encodeURI(url), {
```
binaryoverload, line 162:
```suggestion
    res = await mwFetch<T>(encodeURI(url), {
```

src/backend/metadata/tmdb.ts
```diff
@@ -155,13 +156,32 @@ const headers = {
...
   if (!apiKey) throw new Error("TMDB API key not set");
 
-  const res = await mwFetch<any>(encodeURI(url), {
-    headers,
-    baseURL,
-    params: {
-      ...params,
-    },
-  });
+  let res: Promise<T>;
```
binaryoverload, line 159: We're not returning a promise, we're returning the value itself as the promises are being awaited

```suggestion
  let res: T;
```

src/backend/metadata/tmdb.ts
```diff
@@ -155,13 +156,32 @@ const headers = {
...
+      baseURL,
+      params: {
+        ...params,
+      },
+      signal: AbortSignal.timeout(baseURL !== otherUrl ? 5000 : 30000),
+    });
+  } catch (err) {
+    if (baseURL !== otherUrl) {
+      baseURL = otherUrl;
+      res = await mwFetch<any>(encodeURI(url), {
```
binaryoverload, line 173:
```suggestion
      res = await mwFetch<T>(encodeURI(url), {
```

binaryoverload (changes requested):

src/backend/metadata/tmdb.ts
meghrathod: okay, will do that

src/backend/metadata/tmdb.ts
meghrathod: no, they are not constants in this implementation and I set the baseURL as the other if we do the fallback. but as per @JorrinKievit's suggestion, I'll use boolean for this and implement it so yes then they will be constants

meghrathod:
@JorrinKievit made the required changes, review and let me know

binaryoverload (approved):

ajay01994:
Hi @meghrathod  this  fall back option doesn't work for jio users .it doesn't  load alternate url ,do you any other solution to access TMDB in India ?

binaryoverload:
Please make a new issue for any problems

PR #1121, opened 2024-04-15
Sources reordering

JorrinKievit:
This pull request resolves many user requests.

It still requires a bit of testing.
Also would want to show the name of the source instead of the id in the draggable list.

https://github.com/movie-web/movie-web/assets/43169049/5f380160-7dac-45cc-b0dd-0d4b6fd1065a

 - [ ] I have tested all of my changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
     ]
   },
   "dependencies": {
+    "@dnd-kit/core": "^6.1.0",
+    "@dnd-kit/sortable": "^8.0.0",
+    "@dnd-kit/utilities": "^3.2.2",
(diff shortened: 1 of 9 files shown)


binaryoverload (changes requested):
Works well, good work. 

As discussed internally, I think we need a "reset" order button

Also, if we add/remove sources, this current code doesn't look like it'll handle it as `getProviders` is only called if the settings is empty. If someone has set an order and we delete/add sources, it won't be updated from what I can see?

zisra:
Also how about an embed order?

qtchaos:
> Also how about an embed order?

After an internal discussion, no.

PR #1151, opened 2024-04-22
Add direct bookmarking from movie thumbnails

### LINKED ISSUE #900: [Feature]:

JorrinKievit:
This pull request resolves #900 
I introduced a 'shouldShowBookmark' prop to keep the original edit flow of the bookmarks part on the home page. Otherwise the bookmark button would also show up there, which would counteract the current functionality.

https://github.com/movie-web/movie-web/assets/43169049/afda764d-2f10-4120-9a7d-f8b4a03ac374

### DIFF
--- a/src/components/buttons/IconPatch.tsx
+++ b/src/components/buttons/IconPatch.tsx
@@ -2,7 +2,7 @@ import { Icon, Icons } from "@/components/Icon";
 
 export interface IconPatchProps {
   active?: boolean;
-  onClick?: () => void;
+  onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
   clickable?: boolean;
   className?: string;
   icon: Icons;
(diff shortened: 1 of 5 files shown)


binaryoverload (changes requested):
Looks good!

I'm not sure if this was introduced in this version or not, but when you're editing bookmarks and you remove all of them, the bookmarks section stays in "edit" mode. I think we should make it so that it stops being in edit mode if there are no bookmarks

I'm half inclined to say we don't actually need the edit flow for the bookmarks any more and just use the icon. Reason being that unbookmarking is now twice as hard as bookmarking which I feel is a bit odd

JorrinKievit:
> Looks good!
> 
> I'm not sure if this was introduced in this version or not, but when you're editing bookmarks and you remove all of them, the bookmarks section stays in "edit" mode. I think we should make it so that it stops being in edit mode if there are no bookmarks
> 
> I'm half inclined to say we don't actually need the edit flow for the bookmarks any more and just use the icon. Reason being that unbookmarking is now twice as hard as bookmarking which I feel is a bit odd

Yes this was already an existing issue, but I will fix it in this PR. I'm also down with removing the edit flow for bookmarks, and only keeping it for the continue watching section 👌

Issue #1083, opened 2024-04-03
[Bug]: Custom scrollbar causing layout shifts


zisra:
### What happened?

The custom scrollbar is causing the website's layout to shift on multiple occasions, like opening a modal or searching for something (depends on screen size)

https://github.com/movie-web/movie-web/assets/100528712/f620beb1-c6fe-43b2-ba9c-8f70f9d0d721

sussy-code:
i fixed this on my /discover page using this in my index.css
```css
html,
body {
  @apply bg-background-main font-main text-type-text;
  min-height: 100vh;
  min-height: 100dvh;
  font-size: 1.0235em;
  scrollbar-width: none; /* Hide scrollbar for Firefox */
  -ms-overflow-style: none; /* Hide scrollbar for Internet Explorer and Edge */
}
```
see it [here](https://sudo-flix.lol/discover)

PR #1152, opened 2024-04-22
Fix layout shift

sussy-code:
Fix the layout shifting using a stable [scrollbar gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) and remove it on the player page using the Helmet element.

Original issue: https://github.com/movie-web/movie-web/issues/1083
Discord context: https://discord.com/channels/871713465100816424/871713465578979351/1231997879892054036


zisra:
Does this affect the video player?

sussy-code:
https://streamable.com/0l5po8

here watch it work!

sussy-code:
yea sure, we can do the same helmet thing but id have to add it to all pages who don’t need it, I can do this in a couple hours.

JorrinKievit:
> yea sure, we can do the same helmet thing but id have to add it to all pages who don’t need it, I can do this in a couple hours.

I think the gutter only has to apply on the home page, the rest of the pages we can leave as it is. So you can remove it in the CSS file, and apply the helmet trick to the home page

sussy-code:
fixed

